PR #4, opened 2017-03-17
Expose a method for obtaining a list of available desktop sources

virtuacoplenny:


modules/screensharing/index.js
```diff
@@ -1,25 +1,58 @@
...
 /**
  * Adds screen sharing support. Prompts the user to choose screen, window or
  * tab to share (not implemented yet). And passes the chosen source to caller
  * of the method.
  * @param {Function} callback the success callback
  * @param {Function} errorCallback the callback for errors
  */
-function obtainDesktopStream (callback, errorCallback) {
-    //FIXME: add more types
+function obtainDesktopStream(callback, errorCallback) {
```
bbaldino: should probably have this call `obtainDesktopStreams`?

modules/screensharing/index.js
```diff
@@ -1,25 +1,58 @@
...
+/**
+ * Get sources available for screensharing. The callback is invoked
+ * with an array of DesktopCapturerSources. By default, only 'screen'
+ * sources are obtained unless options.types is specified.
+ *
+ * @param {Function} callback - The success callback.
+ * @param {Function} errorCallback - The callback for errors.
+ * @param {Object} options - Additional configuration for getting sources.
+ * @param {Array} options.types - Specify the desktop source types to get,
+ * with valid sources being "window" and "screen".
```
bbaldino: i'd put the note about defaulting to 'screen' here (instead of at the top)

virtuacoplenny: I ended up removing the default like you had mused about in person.

modules/screensharing/index.js
```diff
@@ -1,25 +1,48 @@
...
-            // first sourceId
-            callback(sources[0].id);
+
+            callback(sources);
         });
 }
 
+/**
+ * Adds screen sharing support. Prompts the user to choose screen, window or
+ * tab to share (not implemented yet). And passes the chosen source to caller
```
saghul: I don't think we support tabs here

modules/screensharing/index.js
```diff
@@ -1,25 +1,48 @@
...
+        (sources) => callback(sources[0].id),
+        errorCallback,
+        { types: ['screen'] }
+    );
+}
+
 module.exports = function setupScreenSharingForWindow(pWindow) {
-    pWindow.JitsiMeetElectron = {obtainDesktopStream};
+    pWindow.JitsiMeetElectron = {
+        obtainDesktopStream,
```
saghul: why expose both?

modules/screensharing/index.js
```diff
@@ -1,25 +1,48 @@
...
- * @param {Function} errorCallback the callback for errors
+ * Get sources available for screensharing. The callback is invoked
+ * with an array of DesktopCapturerSources.
+ *
+ * @param {Function} callback - The success callback.
+ * @param {Function} errorCallback - The callback for errors.
+ * @param {Object} options - Additional configuration for getting sources.
+ * @param {Array} options.types - Specify the desktop source types to get,
+ * with valid sources being "window" and "screen".
+ * @param {Object} options.thumbnailSize - Specify how big the preview
```
saghul: Object? How should that object look like?

modules/screensharing/index.js
virtuacoplenny: The main reason I exposed both was for backwards compatibility. I'm not completely confident on how far to support backwards compatibility between jitsi-meet, lib-jitsi-meet, and jitsi-meet-electron. I wanted to protect interactions between newer and older versions of lib-jitsi-meet with newer and older versions of the electron app. I was also concerned about the case where a custom jitsi-meet implementation does not have an implemented desktop picker, in which case it felt easier to have another method exposed that auto-selects a DesktopCapturerSource--easier because that is the existing implementation that has been working.

saghul: Since the Electron app has never been formally released, I wouldn't worry about backwards compatibility. @hristoterezov WDYT?

saghul: Since nobody objected, can you please drop the old method? I have no further comments other than that. LGTM, great work!

PR #9, opened 2017-04-26
Fix the format of the received remote control events

hristoterezov:
Fixed the format of the received remote control events after the change in https://github.com/jitsi/jitsi-meet/pull/1539#discussion_r113585004


saghul:
does this supersede #7 ?

hristoterezov:
@saghul yes!

PR #10, opened 2017-05-01
Remote control - display the authorization dialog in meet

hristoterezov:
Jitsi Meet part of the PR - https://github.com/jitsi/jitsi-meet/pull/1541

Depends on - https://github.com/jitsi/jitsi-meet-electron/pull/9

### DIFF
--- a/config.js
+++ b/config.js
@@ -1,3 +1,4 @@
 module.exports = {
-    jitsiMeetURL: "https://meet.jit.si/"
+    jitsiMeetURL: "https://meet.jit.si/",
+    handleAuthorization: false
 };
--- a/config.js
+++ b/config.js
@@ -1,4 +1,13 @@
 module.exports = {
+    /**
+     * The URL of the Jitsi Meet deployment that will be used.
+     */
     jitsiMeetURL: "https://meet.jit.si/",
-    handleAuthorization: false
+
+    /**
+     * If true, the authorization of remote control will be handled by jitsi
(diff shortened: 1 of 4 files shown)


modules/remotecontrol/index.js
```diff
@@ -173,16 +181,16 @@ class RemoteControl {
...
-                //Open Dialog and answer
-                this.handlePermissionRequest({
-                    userId: event.userId,
-                    userJID: event.userJID,
-                    displayName: event.displayName,
-                    screenSharing: event.screenSharing
-                });
+                if(event.action === PERMISSIONS_ACTIONS.request
+                    && this.handleAuthorization) {
+                    //Open Dialog and answer
```
saghul: style: space after `//`

config.js
```diff
@@ -1,3 +1,4 @@
 module.exports = {
-    jitsiMeetURL: "https://meet.jit.si/"
+    jitsiMeetURL: "https://meet.jit.si/",
+    handleAuthorization: false
```
saghul: Question: why would someone not want authorization to be handled? Also, why is it false by default?

hristoterezov: After some discussions recently it was decided that Jitsi Meet's default behavior will be to handle the authorization itself. This PR just follows Jitsi Meet's behavior and makes the jitsi-meet-electron project compatible with the latest versions of meet.

modules/remotecontrol/index.js
hristoterezov: Fixed!

PR #12, opened 2017-06-21
make micro mode window load remote video

han-gyeol:
WIP. Don't merge!


config.js
```diff
@@ -2,7 +2,7 @@ module.exports = {
     /**
      * The URL of the Jitsi Meet deployment that will be used.
      */
-    jitsiMeetURL: "https://meet.jit.si/",
+    jitsiMeetURL: "https://meet.jit.si/testurl",
```
bbaldino: probably didn't mean to commit this?

main.js
```diff
@@ -35,6 +46,16 @@ const jitsiMeetWindowOptions = {
 };
 
 /**
+ * Options used when creating the micro mode window.
+ */
+const microWindowOptions = {
```
bbaldino: shouldn't there be something here about it being always on top?

main.js
```diff
@@ -66,17 +87,55 @@ function setAPPListeners () {
  * Opens new window with index.html(Jitsi Meet is loaded in iframe there).
  */
 function createJitsiMeetWindow () {
-    jitsiMeetWindow = new BrowserWindow(jitsiMeetWindowOptions);
-    jitsiMeetWindow.loadURL(indexURL);
+  jitsiMeetWindow = new BrowserWindow(jitsiMeetWindowOptions);
```
bbaldino: it looks like all this got changed from 4 space indent to 2 space indent

main.js
```diff
@@ -66,17 +87,55 @@ function setAPPListeners () {
...
+  jitsiMeetWindow.loadURL(indexURL);
+  microWindow = new BrowserWindow(microWindowOptions);
```
bbaldino: ok, so is this just always loading the micro window for now for testing? (as opposed to only loading it when the main window loses focus?)

main.js
```diff
@@ -66,17 +87,55 @@ function setAPPListeners () {
...
+}
+
+/**
+ * Sets the ipc listeners for messages from renderer processes
+ */
+function setIPCListeners() {
+  // ipcMain.on('log', (event, data) => {
+  //   console.log(data);
+  // });
+  ipcMain.on('mainWindowPeerCreated', () => {
```
bbaldino: i wonder if there might be a better format for these messages where the 'destination' (main window vs micro window) was contained in a field rather than in the message name itself?

windows/jitsi-meet/index.html
```diff
@@ -18,6 +18,7 @@
         </style>
     </head>
     <body>
+      <script src="https://webrtc.github.io/adapter/adapter-latest.js"></script>
```
bbaldino: is this needed by the main window?  i would've thought we were already pulling in adapter in lib-jitsi-meet

windows/jitsi-meet/micro.js
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * The RTCPeerConnection to main window.
+ */
+let peerConnectionMicro;
+
+/**
+ * The remote video from main window.
+ */
+let largeVideo = document.createElement('video');
```
bbaldino: any reason not to just have the video element be defined in the html directly?

windows/jitsi-meet/micro.js
```diff
@@ -0,0 +1,124 @@
...
+    function() {
+      ipcRenderer.send('log', "micro window setLocalDescription complete");
+    },
+    function(error) {
+      ipcRenderer.send('log', "micro window failed to set session description: " + error.toString());
+    }
+  );
+
+  const data = {type: desc.type, sdp: desc.sdp};
+  ipcRenderer.send('microWindowLocalDescriptionSet', data);
```
bbaldino: maybe refactor these calls out into a method like:

  function sendToMainWindow(messageName, data) {
    ipcRenderer.send(messageName, data);
  }

bbaldino: one for the logging calls might be good too?

bbaldino:
made a few comments with some thoughts, probably mostly things you were planning on cleaning up in your refactor anyway.  looking good!

hristoterezov:
@bbaldino Thanks for the comments! I asked @leook0209 to create the PR only for testing purposes. This is not the final version and he is working on fixing it. 


@leook0209 I'll add in the description that this is work in progress. This way everybody will know that this PR is not ready for merge/review.

windows/jitsi-meet/index.html
leook0209: As far as I tried, RTCPeerConnection was undefined without the adapter. I guess lib-jitsi-meet's adapter is contained within the iframe, so the Electron browser does not have access to it.

main.js
leook0209: Currently my plan is to load the micro window when the main window loads, but hidden at first. When the user minimizes /lose focus the main window, I was going to make it visible. For the performance sake, I might gonna make an option to disable micro mode, which prevents the micro window from loading.

main.js
leook0209: My mistake, will be using 4 space indent from now on

windows/jitsi-meet/micro.js
leook0209: Nope, it is for convenience sake for development for now. In the end, all the element will be defined in html.

main.js
```diff
@@ -31,7 +42,19 @@ const jitsiMeetWindowOptions = {
...
-    titleBarStyle: 'hidden'
+    titleBarStyle: 'hidden',
+    webPreferences: { experimentalFeatures: true }
+};
+
+/**
+ * Options used when creating the micro mode window.
+ */
+const microWindowOptions = {
+    width: 800,
```
saghul: we probably want it to be smaller, say 320 x 240. Not resizable.

main.js
```diff
@@ -31,7 +42,19 @@ const jitsiMeetWindowOptions = {
...
+    height: 600,
+    titleBarStyle: 'hidden',
+    frame: true,
```
saghul: I think we should make it frameless. And make the entire window a drag area.

main.js
```diff
@@ -31,7 +42,19 @@ const jitsiMeetWindowOptions = {
...
+    webPreferences: { experimentalFeatures: true }
```
saghul: we don't need this flag for the micro window itself, do we? The capture is done in the main window.

main.js
saghul: I'd like us to test creating the window lazily. Having the window always there will make it "visible" to some window managers on Linux, for example.

main.js
```diff
@@ -66,16 +89,48 @@ function setAPPListeners () {
...
+        if (data) {
+            microWindow.webContents.send(message, data);
+        } else {
+            microWindow.webContents.send(message);
+        }
+    });
+    ipcMain.on('microWindowEvent', (event, args) => {
+        const message = args[0];
+        const data = args[1];
+        if (data) {
```
saghul: why is this needed? doesn't sending undefined work?

modules/micromodeconnection/index.js
```diff
@@ -0,0 +1,260 @@
...
+}
+
+/**
+ * Creates peer connection offer to micro window.
+ */
+function createAndSendOffer() {
+    log("main window createOffer start");
+
+    const offerOptions = {
+        offerToReceiveAudio: 1,
```
saghul: we don't want to receive audio on this one, or the user will hear it twice.

modules/micromodeconnection/index.js
```diff
@@ -0,0 +1,260 @@
...
+ */
+function sendMessageToMicroWindow(message, data) {
+  args = [];
+  args[0] = message;
+  args[1] = data;
+    ipcRenderer.send('mainWindowEvent', args);
+}
+
+
+/*------------------------------Micro Window----------------------------------*/
```
saghul: can we split this file in 2? those globals are not nice to look at :-)

modules/micromodeconnection/index.js
```diff
@@ -0,0 +1,260 @@
...
+    });
+
+    peerConnectionMicro.onicecandidate = function(event) {
+        if(event.candidate !== null) {
+            const newIceCandidate = {
+                candidate: event.candidate.candidate,
+                sdpMLineIndex: event.candidate.sdpMLineIndex,
+                sdpMid: event.candidate.sdpMid
+            };
+            sendMessageToMainWindow('microWindowIceCandidate', newIceCandidate);
```
saghul: why not send the candidate just as it comes? it would also help with signaling the end of candidates.

saghul: Also, we only want to send host candidates.

windows/jitsi-meet/index.html
saghul: We cannot do this. Things may break in different ways. If the Chrome version in Electron doesn't have RTCPeerConnection unprefixed, use webkitRTCPeerConnection please.

windows/jitsi-meet/micro.html
```diff
@@ -0,0 +1,26 @@
...
+        <style>
+            body, html {
+                margin: 0;
+                height: 100%;
+                overflow: hidden;
+            }
+            body {
+                -webkit-app-region: drag
+            }
+            iframe {
```
saghul: why do we need an iframe here?

windows/jitsi-meet/micro.js
saghul: I think you can move this to the HTML now, just query the selector in JS and do the needful there.

windows/jitsi-meet/render.js
```diff
@@ -71,7 +72,16 @@ const dialogFactory = new DialogFactory();
  * Initializes postis.
  * Initializes remote control.
  */
+let largeVideo;
 function onload() {
+    largeVideo = iframe.contentWindow.document.getElementById("largeVideo");
+    // iframe.setAttribute('style', 'width:50%; height:50%; border:0; border:none');
+
+    const canvas = copyVideo(largeVideo, 400, 300);
```
saghul: have you tried to do this without the canvas?

windows/jitsi-meet/render.js
```diff
@@ -85,6 +95,24 @@ function onload() {
 }
 
 /**
+ * Create a copy of HTMLvideo in HTMLcanvas form
+ * Get reference of the video, width and height as parameters
+ * Draw canvas every 33ms
```
saghul: Make this dynamic, so we can change it, we might want to do less than 30 fps.

.eslintrc.js
```diff
@@ -19,19 +19,20 @@ module.exports = {
...
         // While it is considered a best practice to avoid using methods on
         // console in JavaScript that is designed to be executed in the browser
         // and ESLint includes the rule among its set of recommended rules, (1)
         // the general practice is to strip such calls before pushing to
         // production and (2) we prefer to utilize console in lib-jitsi-meet
         // (and jitsi-meet).
         'no-console': 'off',
-        'semi': 'error'
+        'semi': 'error',
+        'no-undef' : 'off'
```
leook0209: Temporarily disabled some lint options because they conflicts with the webkitRTCPeerConnection class.

/home/han/git/jitsi-meet-electron/modules/micromodeconnection/index.js
  49:31  error  A constructor name should not start with a lowercase letter  new-cap
  49:31  error  'webkitRTCPeerConnection' is not defined                     no-undef

Not sure if I should remove them, or find a way around...?

main.js
leook0209: What does it mean by 'lazily'? Is it creating the window when needed instead of at the beginning?

.eslintrc.js
bbaldino: typically there are ways to temporarily disable eslint rules in the files themselves (for exceptional cases), maybe that could be done instead? e.g. https://evanhahn.com/disable-eslint-for-a-file/

main.js
```diff
@@ -31,7 +41,21 @@ const jitsiMeetWindowOptions = {
...
+/**
+ * Options used when creating the micro mode window.
+ */
+const microWindowOptions = {
+    width: 320,
+    height: 240,
+    titleBarStyle: 'hidden',
+    frame: false,
+    alwaysOnTop: true,
+    x: 1570,
```
bbaldino: are these absolute values for the position of the micro window?  wonder if we might get into trouble using absolute values here

main.js
bbaldino: yup, pretty sure that's what saul means...it would definitely be interesting to try that and see how it performs.  if it doesn't perform all that well, maybe we could try some other options like creating the peer connection but only encoding/sending the media when the main window loses focus (to avoid the encoding cost all the time)

saghul: Yep, @bbaldino can read my mind :-)

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "main.js",
   "productName": "Jitsi Meet",
   "scripts": {
-    "start": "electron .",
+    "start": "electron --enable-transparent-visuals --disable-gpu .",
```
leook0209: Temporarily added a gpu disable command for linux (I use Ubuntu 16.04). Not tested yet for other OS

package.json
```diff
@@ -34,9 +34,9 @@
     "postis": "^2.2.0"
   },
   "devDependencies": {
-    "electron": "1.4.13",
+    "electron": "*",
```
leook0209: Installing the latest version of electron doesn't really solve the video.captureStream() problem

windows/jitsi-meet/render.js
```diff
@@ -71,7 +74,21 @@ const dialogFactory = new DialogFactory();
  * Initializes postis.
  * Initializes remote control.
  */
+let largeVideo;
+let mainWindow;
 function onload() {
+    largeVideo = iframe.contentWindow.document.getElementById("largeVideo");
+
+    ipcRenderer.on('hide', () => {
```
leook0209: Lazily create a p2p connection to Micro window

windows/jitsi-meet/render.js
```diff
@@ -71,7 +74,21 @@ const dialogFactory = new DialogFactory();
...
+        setupMicroModePeerConnection(largeVideo.srcObject);
+    });
+
+    largeVideo.ondurationchange = function() {
```
leook0209: this has solved the video transition problem

windows/jitsi-meet/render.js
```diff
@@ -84,6 +101,34 @@ function onload() {
         config.handleRemoteControlAuthorization);
 }
 
+function setupMicroModePeerConnection (stream) {
+    mainWindow = new WindowPeerConnection('jitsiMeetWindow');
+    mainWindow.attachStream(stream);
+    mainWindow.sendStream('microWindow');
+}
+
+function switchMicroVideo (stream) {
```
leook0209: Resend a new MediaStream to Micro window

windows/jitsi-meet/render.js
```diff
@@ -1,23 +1,32 @@
...
-let iframe = document.createElement('iframe');
-iframe.src = process.env.JITSI_MEET_URL || config.jitsiMeetURL;
-iframe.allowFullscreen = true;
-iframe.onload = onload;
-document.body.appendChild(iframe);
+ var options = {
+     domain: config.jitsiMeetURL,
+     room: "testurl",
+     width: '100%',
+     height: '100%'
```
leook0209: It seems like width and height options doesn't work for jitsi-meet ExternalAPI. Regardless of which number I give, it always creates an iframe of MIN width and height, 790 * 300.

windows/jitsi-meet/render.js
```diff
@@ -17,13 +17,15 @@ let channel;
...
-};
-let JitsiMeetAPI = new JitsiMeetExternalAPI(config.jitsiMeetURL, options);
+let domain = config.jitsiMeetURL;
+let room = "testurl";
+let width = '100%';
+let height = '100%';
+let htmlElement = undefined;
+let configOverwrite = {};
+let interfaceConfigOverwrite = {};
+let JitsiMeetAPI = new JitsiMeetExternalAPI(domain, room, width, height,
```
hristoterezov: Why do you need this change?

leook0209: Using options object somehow didnt really work out, some of the parameters were ignored in the iframe created.

leook0209: Oh, if you were asking about using the JitsiMeetExternalAPI, it was necessary for Micro window's toolbar buttons to manipulate the Jtisi-meet application

PR #13, opened 2017-06-29
feat(remotecontrol): multi monitor support

hristoterezov:
Depends on https://github.com/jitsi/jitsi-meet/pull/1694

### DIFF
--- a/config.js
+++ b/config.js
@@ -2,12 +2,5 @@ module.exports = {
     /**
      * The URL of the Jitsi Meet deployment that will be used.
      */
-    jitsiMeetURL: "https://meet.jit.si/",
-
-    /**
-     * If true, the authorization of remote control will be handled by jitsi
-     * meet electron app. Otherwise the authorization will be handled in Jitsi
-     * Meet.
-     */
-    handleRemoteControlAuthorization: false
+    jitsiMeetURL: "https://meet.jit.si/"
 };
--- a/modules/remotecontrol/constants.js
+++ b/modules/remotecontrol/constants.js
@@ -1,8 +1,8 @@
 module.exports = {
     /**
-     * Types of remote-control-event events.
+     * Types of remote-control events.
      */
-    EVENT_TYPES: {
+    EVENTS: {
         mousemove: "mousemove",
         mousedown: "mousedown",
         mouseup: "mouseup",
@@ -10,22 +10,19 @@ module.exports = {
         mousescroll: "mousescroll",
         keydown: "keydown",
         keyup: "keyup",
-        permissions: "permissions",
         stop: "stop",
         supported: "supported"
     },
 
     /**
-     * Actions for the remote control permission events.
+     * Types of remote-control requests.
      */
-    PERMISSIONS_ACTIONS: {
-        request: "request",
-        grant: "grant",
-        deny: "deny"
+    REQUESTS: {
+        start: "start"
     },
 
     /**
-     * The name of remote control events sent trough the API module.
+     * The name of remote control messages.
      */
-    REMOTE_CONTROL_EVENT_NAME: "remote-control-event"
+    REMOTE_CONTROL_MESSAGE_NAME: "remote-control"
 };
--- a/modules/remotecontrol/index.js
+++ b/modules/remotecontrol/index.js
@@ -1,6 +1,8 @@
 let robot = require("robotjs");
+const sourceId2Coordinates = require("sourceId2Coordinates");
+const electron = require("electron");
 const constants = require("../../modules/remotecontrol/constants");
-const {EVENT_TYPES, PERMISSIONS_ACTIONS, REMOTE_CONTROL_EVENT_NAME} = constants;
+const { EVENTS, REQUESTS, REMOTE_CONTROL_MESSAGE_NAME } = constants;
 
 /**
  * Attaching to the window for debug purposes.
@@ -9,13 +11,8 @@ const {EVENT_TYPES, PERMISSIONS_ACTIONS, REMOTE_CONTROL_EVENT_NAME} = constants;
 window.robot = robot;
 
 /**
- * Width/Heught for the screen.
- */
-const {width, height} = robot.getScreenSize();
-
-/**
- * Mouse button mapping between the values in remote-control-event and robotjs
- * methods.
+ * Mouse button mapping between the values in remote control mouse event and
+ * robotjs methods.
  */
 const MOUSE_BUTTONS = {
     1: "left",
@@ -24,8 +21,8 @@ const MOUSE_BUTTONS = {
 };
 
 /**
- * Mouse actions mapping between the values in remote-control-event and robotjs
- * methods.
+ * Mouse actions mapping between the values in remote control mouse event and
+ * robotjs methods.
  */
 const MOUSE_ACTIONS_FROM_EVENT_TYPE = {
     "mousedown": "down",
@@ -33,8 +30,8 @@ const MOUSE_ACTIONS_FROM_EVENT_TYPE = {
 };
 
 /**
- * Key actions mapping between the values in remote-control-event and robotjs
- * methods.
+ * Key actions mapping between the values in remote control key event and
+ * robotjs methods.
  */
 const KEY_ACTIONS_FROM_EVENT_TYPE = {
     "keydown": "down",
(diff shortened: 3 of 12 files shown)


bbaldino:
cool, i learned some new javascript from this pr :)

modules/remotecontrol/constants.js
```diff
@@ -1,31 +1,28 @@
...
         keydown: "keydown",
         keyup: "keyup",
-        permissions: "permissions",
         stop: "stop",
         supported: "supported"
     },
 
     /**
-     * Actions for the remote control permission events.
+     * Types of remote-control requests.
```
virtuacoplenny: So should all instances of "remote control" be hyphenated?  I see comments below, old and new, that are not hyphenated.

modules/remotecontrol/index.js
```diff
@@ -64,136 +61,148 @@ class RemoteControl {
...
-                if(result) {
-                    this.start();
+    start(id, sourceId) {
+        const displays = electron.screen.getAllDisplays();
+
+        switch(displays.length) {
+            case 0:
+                this.display = undefined;
+            break;
+            case 1: // Linux
```
virtuacoplenny: Why is there a comment about Linux here? Is Linux the only case where 1 is returned for display length?

modules/remotecontrol/index.js
```diff
@@ -64,136 +61,148 @@ class RemoteControl {
...
+            // eslint-disable-next-line no-case-declarations
+            default: // > 1 display
+                const coordinates = sourceId2Coordinates(sourceId);
+                if(coordinates) { // Windows
+                    const { x, y } = coordinates;
+                    this.display = electron.screen.getDisplayNearestPoint({
+                        x: x + 1,
+                        y: y + 1
+                    });
+                } else { // ac OS
```
virtuacoplenny: as OS -> Mac OS

modules/remotecontrol/index.js
```diff
@@ -64,136 +61,148 @@ class RemoteControl {
...
+                    // try to match displayID with sourceID
+                    const displayID = Number(sourceId.replace('screen:', ''));
```
virtuacoplenny: Should the capitalization be sourceId or sourceID? I once saw someone change my use of "Id" to "ID."

modules/remotecontrol/index.js
```diff
@@ -64,136 +61,148 @@ class RemoteControl {
...
-     */
-    start() {
-        this.started = true;
+        const response = {
+            id,
+            type: 'response'
+        };
+
+        if(this.display) {
+            this.started = true;
```
virtuacoplenny: I wonder if you could just use this.display to determine started. If you could that would eliminate an instance variable.

modules/remotecontrol/index.js
```diff
@@ -64,136 +61,148 @@ class RemoteControl {
...
+     * Executes the passed message.
+     * @param {Object} message the remote-control message.
      */
-    onRemoteControlEvent(event) {
-        if(!this.started && event.type !== EVENT_TYPES.permissions) {
+    onRemoteControlMessage(message) {
+        const { id, data } = message;
+
+        if(!this.started
+            && data.type != REQUESTS.start) {
```
virtuacoplenny: Could you explain to me why data.type has the start value? Do the messages keep sending that type? Does "start" mean "start doing this action"?

modules/remotecontrol/index.js
```diff
@@ -204,18 +213,31 @@ class RemoteControl {
...
         this.channel.send({
             method: 'message',
             params: { data: remoteControlEvent }
         });
     }
+
+    /**
+     * Sends a response of a request from Jitsi Meet.
+     *
+     * @param {Object} response the response.
```
virtuacoplenny: Could you expand on this comment and say what the response is from?

config.js
```diff
@@ -2,12 +2,5 @@ module.exports = {
...
      * The URL of the Jitsi Meet deployment that will be used.
      */
-    jitsiMeetURL: "https://meet.jit.si/",
-
-    /**
-     * If true, the authorization of remote control will be handled by jitsi
-     * meet electron app. Otherwise the authorization will be handled in Jitsi
-     * Meet.
-     */
-    handleRemoteControlAuthorization: false
```
virtuacoplenny: Why did you decide to remove this?

modules/remotecontrol/constants.js
hristoterezov: No. Will fix it!

modules/remotecontrol/index.js
hristoterezov: Do you mean that I have to change it to - "the response of request"?

config.js
hristoterezov: Because the next thing we have to do is to make sure that the user is sharing only entire screen (not an app window or tab). This will change the message of the dialog and we'll have to pass more arguments. I felt that actually supporting two equivalent windows (here and in meet) is becoming hard to maintain. 

Also I'm not sure if anybody will want to use this dialog ever and we need to make it prettier probably. 

That's why I decided to remove it for now and at some point if we need such a dialog we can revert it.

modules/remotecontrol/index.js
hristoterezov: Good idea! Will check that!

modules/remotecontrol/index.js
virtuacoplenny: You don't have to change it if you don't want but seeing a doc that says "response the response" seems unnecessary or at least redundant. A suggestion might be to say what generated the response or what the response will be used for.

modules/remotecontrol/index.js
hristoterezov: This is the remote control message format coming from meet. You can check modules/transport in Jitsi Meet project. What happens is that Transport class implements sending requests. The format for the requests is `{id, type: 'request', data}`. `data` prop is defined by every consumer that uses the Transport module. The remote control module in Jitsi Meet project uses this field too. 

All remote control messages have `data.name = 'remote-control'`. This way the electron app can filter them from all other messages. Also all remote control messages are defining `data.type` to specify an operation to be executed by the electron app.

In this case we have `data.type = 'start'` which we are using to notify the electron app that a remote control session will start and to pass some initial data that will be needed( like the desktop sharing source id) by the electron app.

Do you have any suggestions for improvement? Maybe renaming?

modules/remotecontrol/index.js
hristoterezov: Will check that!

modules/remotecontrol/index.js
virtuacoplenny: Nope, that was a great explanation. I get it now... If remote control hasn't started yet, always wait for the start request before doing anything. It's plainly obvious now.

modules/remotecontrol/index.js
hristoterezov: No :) I'll expand the comment here! What I have observed for now is that on linux even if you have multiple monitors, they are concatenated in 1 big screen and here the length of displays array will always be 1.

modules/remotecontrol/index.js
hristoterezov: I changed the `sourceID` and `displayID` to `sourceId` and `displayId` here. In Jitsi Meet and lib-jitsi-meet is used only `sourceId`. This was the easiest fix.

But if you insist on `sourceID` and `displayID` and you think the current one is wrong I'll change it. But I'll have to change it for Jitsi Meet and lib-jitsi-meet too.

modules/remotecontrol/index.js
hristoterezov: Sure! I agree that looks strange! Will add more explanations there.

hristoterezov: You are totally right! When I looked how to change it, I realized that the method actually only sends a message :) It can be easily used to send any message not only responses! So I renamed the method. I hope the new version is clearer!

hristoterezov:
@virtuacoplenny I think I addressed all your comments! Could you please check the PR again?

modules/remotecontrol/index.js
virtuacoplenny: I don't insist. I don't care too much either way, although I am a fan of consistency. I'm just bringing it up because someone changed my previous code to be capital "ID" and I don't know if there's an undocumented rule about that.

config.js
virtuacoplenny: Should "requestRemoteControlPermissions" be cleaned up from jitsi-meet then?

modules/remotecontrol/index.js
```diff
@@ -50,150 +47,163 @@ const KEY_ACTIONS_FROM_EVENT_TYPE = {
...
-     * Construcs new instance.
-     * @constructor
-     */
-    constructor() {
-        this.started = false;
-    }
-
-    /**
-     * Initializes the remote control functionality.
+     * Construcs new instance and initializes the remote control functionality.
```
virtuacoplenny: Construcs -> Constructs

modules/remotecontrol/index.js
```diff
@@ -50,150 +47,163 @@ const KEY_ACTIONS_FROM_EVENT_TYPE = {
...
+        };
+
+        if(this.display) {
+            response.result = true;
+        } else {
+            response.error
+                = 'Error: Can\'t detect the display that is currently shared';
+        }
+
+        this.sendResponse(response);
```
virtuacoplenny: I can't find where sendResponse is defined.

modules/remotecontrol/index.js
hristoterezov: Fixed!

modules/remotecontrol/index.js
hristoterezov: Oh! Sorry for this! Fixed!

PR #14, opened 2017-07-07
Extract utils

hristoterezov:


virtuacoplenny:
What type of logic should be left in jitsi-meet-electron going forward?

virtuacoplenny:
I'm also thinking maybe an addition to the readme about how to develop the utils library locally would be helpful. I think jitsi-meet has that about making changes to lib-jitsi-meet and I think it's a useful reference--I look at the jitsi-meet README each time I want to make lib-jitsi-meet changes.

hristoterezov:
@virtuacoplenny:

> What type of logic should be left in jitsi-meet-electron going forward?

I don't understand the question! Probably everything that used only in jitsi-meet-electron ...

> I'm also thinking maybe an addition to the readme about how to develop the utils library locally would be helpful. I think jitsi-meet has that about making changes to lib-jitsi-meet and I think it's a useful reference--I look at the jitsi-meet README each time I want to make lib-jitsi-meet changes.

I'm not sure what are you proposing. Are you talking about:
https://github.com/jitsi/jitsi-meet#working-with-the-library-sources-lib-jitsi-meet

virtuacoplenny:
The two feautures of the electron app were getting screensharing sources and remote control. That logic was moved out of jitsi-meet-electron. So the question is what is the purpose of jitsi-meet-electron? Should all future feature work go into the utils repo?

And yes, I'm refereing to adding a section like https://github.com/jitsi/jitsi-meet#working-with-the-library-sources-lib-jitsi-meet to the jitsi-meet-electron repo. At this point my concern is someone wanting to create a new feaute and wondering how to go about that.

saghul:
> The two feautures of the electron app were getting screensharing sources and remote control. That logic was moved out of jitsi-meet-electron. So the question is what is the purpose of jitsi-meet-electron? Should all future feature work go into the utils repo?

I'd say this repo is the showcase of those features. The utils repo makes it possible to integrate them into other projects, and this project is a simple "barebones" project which has nothing more, other than Jitsi Meet. That's how I see it :-)

hristoterezov:
I agree with @saghul ! But also this is the case now, maybe in future we're going to put something more in jitsi-meet-electron project. For now we don't a have a rule "everything should go into utils repo" or at least I'm not aware :), if this is what you are asking!

PR #15, opened 2017-07-14
Use iframe API

hristoterezov:


windows/jitsi-meet/render.js
```diff
@@ -1,40 +1,35 @@
...
+    document.head.appendChild(script);
 }
 
-/**
- * Clears the postis objects and remoteControl.
- */
-function onunload() {
-    remoteControl.dispose();
-}
+loadScript(`https://${jitsiMeetDomain}/external_api.js`, () => {
```
virtuacoplenny: Should the external_api.js domain also be configurable? I maybe could see a use case where I want to test my api changes in electron. But I haven't done enough work with electron/iframe-api to know if it'll be useful.

windows/jitsi-meet/render.js
```diff
@@ -1,40 +1,35 @@
...
+    script.src = src;
+    document.head.appendChild(script);
 }
 
-/**
- * Clears the postis objects and remoteControl.
- */
-function onunload() {
-    remoteControl.dispose();
-}
```
virtuacoplenny: Why was remoteControl.dispose() needed before and not needed now?

virtuacoplenny:
Relies on https://github.com/jitsi/jitsi-meet-electron-utils/pull/2

config.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
     /**
-     * The URL of the Jitsi Meet deployment that will be used.
+     * The domain of the Jitsi Meet deployment that will be used.
      */
-    jitsiMeetURL: "https://meet.jit.si/"
+    jitsiMeetDomain: "meet.jit.si"
```
virtuacoplenny: You'll need to update the README for this change.

windows/jitsi-meet/render.js
hristoterezov: Exactly!

windows/jitsi-meet/render.js
hristoterezov: This way we make sure that the script will be loaded from the same deployment specified in the config.js. 

For me it is very useful for testing. But I think it will be useful in general too.

If somebody has it's own deployment and want to use the app with it probably will want to load the external_api.js from the same place.

Do you have any other suggestions? We may want to add another config option only for the script but I think this is a little bit too much for now? Or we can put 'meet.jit.si' there but if somebody want to change it, I think it won't be very convenient?

config.js
hristoterezov: Fixed!

PR #23, opened 2018-05-07
Updated Electron to 2.0.0

AkshitKrNagpal:


AkshitKrNagpal:
@saghul 
Can you merge this? I wasn't merged after being reviewed.

saghul:
@AkshitKrNagpal Merged, thanks! I amended your patch to pin all dependencies, otherwise the lock file is not that useful :-)

PR #25, opened 2018-05-09
Used eslint-config-jitsi and modified code

AkshitKrNagpal:


AkshitKrNagpal:
@saghul 
For `jitsi-meet-electron/main.js`
`61:9   warning  Arrow function has too many parameters (6). Maximum allowed is 3  max-params`
I don't think anything can be done. The `certificate-error` event returns 6 params [link](https://github.com/electron/electron/blob/master/docs/api/app.md#event-certificate-error)

`61:30  warning  'url' is already declared in the upper scope                      no-shadow
  79:58  warning  'url' is already declared in the upper scope                      no-shadow`
I can rename upper scope `url` to `URL`

For `jitsi-meet-electron/windows/jitsi-meet/render.js`
This file will be fully replaced after react configuration. We can ignore this for now.

PR #26, opened 2018-05-09
Fix no-shadow warning

AkshitKrNagpal:


.eslintrc.js
```diff
@@ -4,7 +4,6 @@ module.exports = {
     ],
     'rules': {
         'no-new': 1,
-        'no-shadow': 1,
```
saghul: why are we removing the rule? is it already covered by the jitsi rules?

AkshitKrNagpal: These rules are defined as errors in jitsi config.
I changed them to warnings so that I can commit these changes.
Since no-shadow is fixed, there is no need for that.

saghul: Thanks!

PR #28, opened 2018-05-11
Configured React using Webpack with current functionality

### LINKED ISSUE #21: No Webpack (or React)
It will be nice to have Webpack to generate static assets for the application.
It would be beneficial to create a welcome screen or any other offline components.

I can work on it.

AkshitKrNagpal:
Fixes #21


webpack.config.js
```diff
@@ -0,0 +1,53 @@
...
+        entry: { main: './main.js' }
+    },
+    commonConfig),
+    Object.assign({
+        target: 'electron-renderer',
+        entry: { renderer: './windows/jitsi-meet/src/index.js' },
+        plugins: [
+            new HtmlWebpackPlugin({
+                title: 'Jitsi Meet',
+                template: path.resolve('./templates/react-template.html')
```
saghul: Can you please rename this to "index.html" instead?

AkshitKrNagpal: Renamed. Squashed. Updated.

windows/jitsi-meet/src/index.js
```diff
@@ -0,0 +1,57 @@
...
+     * @return {ScriptElement}
+     */
+    render() {
+        return null;
+    }
+
+    /**
+     * Attach the script
+     */
+    componentDidMount() {
```
saghul: Let's do it on WillMount.

windows/jitsi-meet/src/index.js
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Jitsi Meet Window Component
+ */
+class JitsiMeetWindow extends React.Component {
+
+    /**
+     * Render function of component
+     * @return {ScriptElement}
+     */
+    render() {
```
saghul: We don't need the render method at all.

windows/jitsi-meet/src/index.js
```diff
@@ -0,0 +1,57 @@
...
+        script.onerror = console.error;
+        script.src = `https://${jitsiMeetDomain}/external_api.js`;
+
+        document.head.appendChild(script);
+    }
+
+    /**
+     * When the script is loaded attach utils from jitsi-meet-electron-utils
+     */
+    onScriptLoad() {
```
saghul: Make this method "private", prefix it with `_`, that is.

windows/jitsi-meet/src/index.js
AkshitKrNagpal: Not defining render method will give an error. Because react executes it.
render

windows/jitsi-meet/src/index.js
AkshitKrNagpal: Are you sure?

**componentDidMount()**
https://reactjs.org/docs/react-component.html#componentdidmount
"If you need to load data from a remote endpoint, this is a good place to instantiate the network request."
Not the same thing. But similar maybe.

windows/jitsi-meet/src/index.js
saghul: Doh, I thought it would just return null by default, sigh.

windows/jitsi-meet/src/index.js
saghul: You're right, let's keep it there. Later, when we add some form of loading indicator.

windows/jitsi-meet/src/index.js
saghul: We don't follow that style guide. We follow our own :-) See the jitsi-meet project, we use it extensively.

saghul: So yeah, please change it.

AkshitKrNagpal: Yeah. I know we don't follow that guide. But their reason seems good.
Adding "_" as the prefix neither does anything nor it is a convention. 

Anyways, I have updated the PR.

PR #29, opened 2018-05-21
[RFC] Reorganize source code

saghul:


src/features/app/index.js
```diff
@@ -0,0 +1 @@
+export * from './components';
```
AkshitKrNagpal: Why make a folder in features for App?
We can just place app.js with ./src/index.js.
Components folder seems unnecessary. 
WDYT?

src/features/app/components/App.js
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Main component encapsulating the entire application.
+ */
+export default class App extends Component {
+    /**
+     * Initializes the application title.
+     *
+     * @inheritdoc
+     */
+    componentWillMount() {
```
AkshitKrNagpal: I think we should not use componentWillMount(). It is being deprecated.
They have changed its name to [UNSAFE_componentWillMount()](https://reactjs.org/docs/react-component.html#unsafe_componentwillmount
) in latest React versions (with backwards compatibility)

https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html

src/features/config/index.js
```diff
@@ -0,0 +1,12 @@
...
+    /**
+     * Application name.
+     */
+    appName: 'Jitsi Meet',
+
+    /**
+     * The domain of the Jitsi Meet deployment that will be used.
+     */
+    defaultDomain: 'meet.jit.si'
+};
```
AkshitKrNagpal: Maybe this file should be better as src/config/index.js.
WDYT?

src/features/app/index.js
saghul: Because it keeps the structure consistent and other features will have multiple components. Also extensibility. If we want to add more components later, I don't want to keep changing the structure.

src/features/app/components/App.js
saghul: Good point. I'll move it to the constructor instead, since in this case all we do is change the document.

src/features/config/index.js
saghul: I'm following the same pattern as jitsi-meet here. Our idea of a feature is an abstraction of certain functionality, which other features can then use / consume. In that line, the "config" feature exposes some defaults to other features. Also, when redux is added, this will grow to store the user set display name, server URL, etc.

src/features/app/index.js
AkshitKrNagpal: Okay cool.!

src/features/config/index.js
AkshitKrNagpal: Hmm... I see.

AkshitKrNagpal:
Looks good. 👍

saghul:
@AkshitKrNagpal can you please look into fixing the packing, with this PR? I couldn't figure out how to solve it... seems like the `__dirname` thing again. Thanks!

AkshitKrNagpal:
Okay. Let me check it out.

AkshitKrNagpal:
Can you give me steps to reproduce error?

saghul:
> Can you give me steps to reproduce error?

Pack the app with `npm run pack` and then try to open the resulting application.

AkshitKrNagpal:
`main` in `package.json` has to be `build/main.js` (I think you were experiencing that error?).
But `__dirname` will be still `/` so HTML file is not loading. For packaged app I think we'll need to use `electron.app.getAppPath()` because I can't really find the reason behind `__dirname` to be `/`. I tested it and it will work. We'll need to check environment then decide which one to use. WDYT?

I found that some people experience problems with `react-router` and `electron` so just bookmarked this for future. https://github.com/sindresorhus/electron-serve 
This might solve both for us. Can you take a look @saghul ?

AkshitKrNagpal:
@saghul 
Also, I think we can still merge this. I will send a new pull request with Webpack build configuration and pack when we decide what approach to take.

saghul:
Also, @gpolitis I have been forgetting you 🤦‍♂️

hristoterezov:
@saghul Isn't it better to land it when we also have the fix for packaging?

hristoterezov:
@saghul I'd keep main.js separate from all render stuff. Also maybe I'd be in favor of moving the html file somewhere.

AkshitKrNagpal:
@hristoterezov 
This package uses that only. See - [link](https://github.com/sindresorhus/electron-serve/blob/d2695d12104e55dfb884e3ec7be5758bf07ebca6/index.js#L34)
Plus it registers a new protocol (using electron API). 

Sure we can try to find what others are doing.

AkshitKrNagpal:
> I'd keep main.js separate from all render stuff. Also maybe I'd be in favor of moving the html file somewhere.

@hristoterezov @saghul 
How about creating `main` and `renderer` folders in `src` ?
We can move `index.html + index.js + features` to `renderer` and `main.js` to `main`
That way we have a structure for main process files. ( like autoupdater maybe? )

Just a thought. For `index.html`, since it is technically a template we can also make `templates` folder in `src` for it.

hristoterezov:
@AkshitKrNagpal I'd like a more standard solution. If it turns out there isn't one maybe we can use it.

saghul:
> I'd keep main.js separate from all render stuff. Also maybe I'd be in favor of moving the html file somewhere.

I can move main.js out, and maybe rename src/ to app/, so everything in app/ is the web stuff and the main.js out is the Electron stuff. You like?

saghul:
> How about creating main and renderer folders in src ?

I don't really like that. The logical separation between what is the web app and what isn't sounds better to me.

> For index.html, since it is technically a template we can also make templates folder in src for it.

I think we can even avoid making it a template since we know what the generated file will be. I've also seen projects use web-acp-dev-server in dev mode instead of the bundle, so that can also be interesting, and this was accomplished with some conditions in the `<script>` in the HTML.

saghul:
@hristoterezov How about this: I'll make a couple more changes. Then we can merge this into some `dev` branch maybe, where @AkshitKrNagpal can continue and fix the packaging thing.

@AkshitKrNagpal WDYT?

AkshitKrNagpal:
Cool. 

But we can actually avoid a dev branch. In package.json, set main to main.js (currently in src folder).
Packaged app will work. So I think we can merge this to master. 

Then I'll make the changes.

saghul:
Thanks @AkshitKrNagpal, I'll give it a try! Have you tested that?

AkshitKrNagpal:
@saghul 
Yes, I have. The production will then use the original `main.js`.
Once this is merged I will configure build script for webpack production build and maybe webpack-dev-server.
Then I can make modifications for the packaged app to use webpacked `main.js`. (if we decide that)

Is this fine?

saghul:
@AkshitKrNagpal you're the man! Yes it works!

@hristoterezov can you please approve / merge now?

hristoterezov:
> I can move main.js out, and maybe rename src/ to app/, so everything in app/ is the web stuff and the main.js out is the Electron stuff. You like?

👍 Let's do this!

saghul:
That’s what this PR does now, minus the app folder rename.

saghul:
@hristoterezov Renamed src/ to app/

PR #31, opened 2018-05-24
Configured flow

### LINKED ISSUE #30: Start using Flow

AkshitKrNagpal:
@saghul 
Fixes #30 

Comments?
I believe I have to configure [babel-plugin-react-flow-props-to-prop-types](https://github.com/atlassian/babel-plugin-react-flow-props-to-prop-types) next to make it working?
Am I going in right direction?

PS - I am flowing these [docs](https://flow.org/en/docs/react/).


app/features/app/components/App.js
```diff
@@ -1,12 +1,14 @@
-import React, { Component } from 'react';
+// @flow
+
+import * as React from 'react';
```
saghul: leave this as it was, please. See what we do in jitsi-meet/react for reference.

app/features/conference/components/Conference.js
```diff
@@ -24,6 +28,7 @@ export default class Conference extends Component {
         script.onerror = console.error;
         script.src = `https://${config.defaultDomain}/external_api.js`;
 
+        // $FlowFixMe
```
saghul: why is this one needed?

app/index.js
```diff
@@ -1,9 +1,12 @@
...
+// @flow
+
+import * as React from 'react';
 import { render } from 'react-dom';
 
 import { App } from './features/app';
 
 /**
  * Render the main / root application.
+ * $FlowFixMe
```
saghul: same

package.json
```diff
@@ -39,6 +41,7 @@
     "babel-eslint": "8.2.3",
     "babel-loader": "7.1.4",
     "babel-preset-env": "1.7.0",
+    "babel-preset-flow": "6.23.0",
```
saghul: let's leave this out for later. it strips type information, which we don't necessarily want to do.

AkshitKrNagpal: okay.. Cool. I'll remove this.

app/features/conference/components/Conference.js
AkshitKrNagpal: It was taking head as null and saying appendChild is not function of null

AkshitKrNagpal: > Cannot call document.head.appendChild because property appendChild is missing in
> null [1].

app/index.js
AkshitKrNagpal: > Cannot call render with document.getElementById(...) bound to container because
> null [1] is incompatible with Element [2].

app/features/app/components/App.js
saghul: Types can be imported separately, which we prefer:

```
import React, { Component, type Node } from 'react';
```

> I couldn't find // @flow comment in jitsi-meet/react folder. How are you using flow there?

We use it, a lot: https://github.com/jitsi/jitsi-meet/search?utf8=%E2%9C%93&q=%22%40flow%22&type=

app/features/conference/components/Conference.js
saghul: 👍

app/index.js
saghul: 👍

app/features/app/components/App.js
AkshitKrNagpal: > Types can be imported separately, which we prefer:
> import React, { Component, type Node } from 'react';

Cool. I'll use it like this then.

> We use it, a lot:

They are not in all files. When do you use it and when you don't?

AkshitKrNagpal:
Ping @saghul 

Also,

> I believe I have to configure babel-plugin-react-flow-props-to-prop-types next to make it working?

Should I do this in this pull request only?

saghul:
> I believe I have to configure babel-plugin-react-flow-props-to-prop-types next to make it working?
> 
> Should I do this in this pull request only?

Do you currently get an error? Also, can you please rebase it?

AkshitKrNagpal:
No, currently there is no error. Because we haven't defined any state or props which need flow to prop-types conversion.

This can also be configured later when we define props for our components later. 

I'll rebase it.

AkshitKrNagpal:
Ping @saghul

PR #33, opened 2018-05-30
Created basic Welcome page

AkshitKrNagpal:


saghul:
@AkshitKrNagpal wow, amazing work here! I just tested it and it works great! I'll start reviewing the code now.

app/features/welcome/components/Welcome.js
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    /**
+     * Render function of component.
+     *
+     * @return {ReactElement}
+     */
+    render() {
+        return (
+            <AtlasKitThemeProvider mode='light'>
```
saghul: why don't use the theme provider in the root component instead of directly in the welcome page? this way, whatever other routes we add would use the same theme.

app/features/welcome/components/Welcome.js
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    /**
+     * Keeps URL input value and URL in state in sync
+     */
+    _onURLChange(event: Event) {
+        this.setState({
+
+            // $FlowFixMe
+            url: event.target.value
```
saghul: Is there no flow type definition for this?

app/features/welcome/components/Welcome.js
```diff
@@ -0,0 +1,145 @@
...
+    _onFormSubmit(event: Event) {
+        event.preventDefault();
+        this._onJoin();
+    }
+
+    /**
+     * Redirect and join conference
+     */
+    _onJoin() {
+
```
saghul: style: remove empty line

app/features/welcome/components/Welcome.js
```diff
@@ -0,0 +1,145 @@
...
+        const url = URL.parse(this.state.url);
```
saghul: I guess this could give an exception, we need to catch it.

app/features/welcome/components/Welcome.js
```diff
@@ -0,0 +1,145 @@
...
+
+        // Check if the parsed url is a full url or just room name
+        if (url.host) {
```
saghul: You also need to check if there is a path different than `/` and not do anything in that case.

saghul:
I have not commented on the welcome page style, let's discuss that once the inner works of this PR are complete.

app/features/welcome/components/Welcome.js
```diff
@@ -0,0 +1,145 @@
...
+                        <form onSubmit = { this._onFormSubmit }>
+                            <FieldTextStateless
+                                autoFocus = {true}
+                                isLabelHidden = {true}
+                                shouldFitContainer = { true }
+                                onChange = { this._onURLChange }
+                                type = 'text'
+                                value = { this.state.url } />
+                        </form>
+                        <Button
```
saghul: This button should be disabled if the URL is not valid.

app/features/welcome/components/Welcome.js
```diff
@@ -0,0 +1,145 @@
...
+                            appearance='primary'
+                            type='button'
```
saghul: style: space between =

app/features/app/components/App.js
```diff
@@ -3,8 +3,23 @@
 import React, { Component } from 'react';
 
 import { Conference } from '../../conference';
+import { Welcome } from '../../welcome';
+
 import config from '../../config';
 
+// $FlowFixMe
+import styled from 'styled-components';
```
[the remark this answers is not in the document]
AkshitKrNagpal: Cool. I'll do this.

app/features/app/components/App.js
```diff
@@ -3,8 +3,23 @@
...
+
+const Container = styled.div`
```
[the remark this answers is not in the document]
AkshitKrNagpal: Yeah. I just placed them there so I can check if they work. 
They definitely need some organisation.

> They have styled components in a styled.js file or in individual files inside a styled directory if need be. I think we can just go with the single file inside each feature.

It will be good either way. But I think we should go with directory. They will be much more organised and will be better if we have more number of styled components. (Thinking about future too)

> When a component "augments" the styled thing, the styled object is called FooElement and the component is called Foo, which renders a FooElement, among doing other things.

I actually think this has benefits too.
But they are same as react components which can be placed in `components` directory of `features`. If we need styled components in it. we can import from `styled` directory.

app/features/app/components/App.js
```diff
@@ -3,8 +3,23 @@
...
+    height: 100vh;
+`;
+
+// $FlowFixMe
+import { AtlasKitThemeProvider } from '@atlaskit/theme';
+
+// $FlowFixMe
+import { HashRouter as Router, Route } from 'react-router-dom';
```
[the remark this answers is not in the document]
AkshitKrNagpal: I think you meant Alpha [Repo link](https://github.com/ReactTraining/react-router/tree/master/packages/react-router-redux)

**My thoughts**
I personally stay away for alpha versions. Betas are okay.
But yeah app is not much complex. I don't think we will run into any problems.
Lets try when we have redux setup.

BTW. I planned to configure redux and modify code in next pull request.

app/features/conference/components/Conference.js
```diff
@@ -17,43 +17,92 @@ import config from '../../config';
  * Conference component.
  */
 export default class Conference extends Component<*> {
+
+    /**
+     * It will be the reference to the element of this component.
+     */
+    _currentNode: Object;
```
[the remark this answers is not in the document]
AkshitKrNagpal: Sure. I'll change it.

app/features/conference/components/Conference.js
```diff
@@ -17,43 +17,92 @@ import config from '../../config';
...
+    _api: Object;
+
+    /**
+     * Initializes a new {@code Conference} instance.
+     *
+     * @inheritdoc
+     */
+    constructor() {
+        super();
+        this._currentNode = React.createRef();
```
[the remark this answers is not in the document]
AkshitKrNagpal: Ok.

For all `style:` comments - Are you checking with any tool?

app/features/conference/components/Conference.js
```diff
@@ -17,43 +17,92 @@ import config from '../../config';
...
-     * Render function of component.
+     * Implements React's {@link Component#render()}.
      *
-     * @return {ReactElement}
+     * @inheritdoc
+     * @returns {ReactElement}
      */
     render() {
-        return null;
+        return <div ref={this._currentNode} style= {{ height: '100vh' }}/>;
```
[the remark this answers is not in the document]
AkshitKrNagpal: Yeah. I'll reorganise it.

app/features/conference/components/Conference.js
```diff
@@ -17,43 +17,92 @@ import config from '../../config';
...
+        if (this._api) {
+            this._api.dispose();
+        }
+    }
+
+    /**
+     * Navigates to home screen (Welcome Screen)
+     */
+    _navigateToHome() {
+        this.props.history.push('/');
```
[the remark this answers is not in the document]
AkshitKrNagpal: Sure

app/features/welcome/components/Welcome.js
```diff
@@ -0,0 +1,145 @@
...
+        super(props);
+
+        this.state = {
+            url: ''
+        };
+
+        // Bind event handlers
+
+        // $FlowFixMe
+        this._onURLChange = this._onURLChange.bind(this);
```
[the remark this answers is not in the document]
AkshitKrNagpal: I searched this and tried something like it. 
It didn't work. Maybe I did something wrong.
I'll try again and let you know.

app/features/welcome/components/Welcome.js
AkshitKrNagpal: We planned on using the dark theme in the previous meeting. So I did this.

jitsi-meet also does this. It uses light theme for welcome page only. If we use light for whole app. It shows white background during loading of conference. Dark theme makes it look good.

https://github.com/jitsi/jitsi-meet/blob/ce69ee60ca6b1eb6210e38bc3cc68cee60d42b15/react/features/app/components/App.web.js#L53

https://github.com/jitsi/jitsi-meet/blob/ce69ee60ca6b1eb6210e38bc3cc68cee60d42b15/react/features/welcome/components/WelcomePage.web.js#L106

app/features/welcome/components/Welcome.js
AkshitKrNagpal: I tried a bunch of domains and room cases. It worked as expected. 
But yeah [docs](https://nodejs.org/api/url.html#url_url_parse_urlstring_parsequerystring_slashesdenotehost) says it can throw error. I'll catch them.

app/features/welcome/components/Welcome.js
saghul: Got it, makes sense!

app/features/conference/components/Conference.js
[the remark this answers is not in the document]
saghul: My Human Eyes (R) 😄

app/features/app/components/App.js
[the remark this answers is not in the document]
saghul: Makes sense, let's do this!

app/features/welcome/components/Welcome.js
AkshitKrNagpal: If path is `/` then it redirects to `/` so stays on the same page.

app/features/app/components/App.js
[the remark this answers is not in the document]
saghul: Sounds great!

app/features/welcome/components/Welcome.js
AkshitKrNagpal: Yeah... 
I have modified welcome page more and trying to do something like in the mobile app. using [onboarding](https://atlaskit.atlassian.com/packages/core/onboarding) component (I'll push that too when I'm done)

But there is some bug in that. I couldn't figure out if I am doing something wrong or component is behaving wrongly. 

I'll update you with this.

app/features/conference/components/Conference.js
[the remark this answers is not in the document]
AkshitKrNagpal: Yeah... They are the best. 😁😂

app/features/welcome/components/Welcome.js
AkshitKrNagpal: Found one. Will try and update this.

app/features/welcome/components/Welcome.js
saghul: Yes, but I think we shouldn't enable the button when the URL is invalid. Also, I guess we should check if the scheme is http or https, otherwise it's not a valid URL!

app/features/welcome/components/Welcome.js
saghul: Let's leave onboarding for later. For now, if the URL is not correct, we should not do anything if the user taps enter or the GO button.

app/features/welcome/components/Welcome.js
AkshitKrNagpal: What if the user doesn't enter protocol?
Let's say it enters 'beta.meet.jit.si/room'. Shouldn't it go to `beta.meet.jit.si` server and join room `room`??

saghul: For consistency with the mobile app: nope. I know it makes sense to try to do that, so maybe we can revisit this in the future.

app/features/app/components/App.js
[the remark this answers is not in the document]
AkshitKrNagpal: Wait.
I am already ignoring `node_modules` folder. See [link](https://github.com/jitsi/jitsi-meet-electron/blob/6f35ddd792c5a3e9cc462c31cb1166fa1fbbdb09/.flowconfig#L2).

`$FlowFixMe` is for this error.

screen shot 2018-05-31 at 4 42 03 pm

Any ideas around this?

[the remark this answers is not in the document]
AkshitKrNagpal: Ping @saghul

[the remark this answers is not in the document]
saghul: Hum. Maybe that's it. You are ignoring all modules, so it will not find them, maybe?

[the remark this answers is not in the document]
saghul: Hum. I see. Let's table it for now. That is, leave it as is.

app/features/app/components/App.js
```diff
@@ -28,7 +43,16 @@ export default class App extends Component<*> {
      */
     render() {
         return (
-            <Conference />
+            <AtlasKitThemeProvider mode = 'dark'>
+                <Router>
+                    <Container>
+                        <Route exact path = '/' component = {Welcome} />
```
saghul: style: spaces around braces: `{ Welcome }`.

app/features/conference/components/Conference.js
```diff
@@ -13,47 +13,109 @@ import {
 
 import config from '../../config';
 
+type Props = {
+
+    /**
+     * React Router history object (for flow).
```
saghul: can you be a bit more descriptive here? also, this has nothing to do with flow.

app/features/conference/components/Conference.js
```diff
@@ -13,47 +13,109 @@ import {
...
+     */
+    history: Object;
+
+    /**
+     * Match Props Params.
+     */
+    match: Object;
```
saghul: ditto.

app/features/conference/components/Conference.js
```diff
@@ -13,47 +13,109 @@ import {
...
+};
+
 /**
  * Conference component.
  */
-export default class Conference extends Component<*> {
+export default class Conference extends Component<Props, *> {
+
```
saghul: style: remove extra newline

app/features/conference/components/Conference.js
```diff
@@ -13,47 +13,109 @@ import {
...
-     * Render function of component.
+     * Implements React's {@link Component#render()}.
      *
-     * @return {ReactElement}
+     * @inheritdoc
+     * @returns {ReactElement}
      */
     render() {
-        return null;
+        return <div ref={this._ref} style={{ height: '100vh' }} />;
```
saghul: use a `Container` here too.

app/features/welcome/components/Welcome.js
```diff
@@ -0,0 +1,143 @@
...
+    justify-content: center;
+`;
+
+const WelcomeWrapper = CenterContent.extend`
+    background: linear-gradient(#165ecc,#44A5FF);
+    height: 100vh;
+    width: 100%;
+`;
+
+// $FlowFixMe
```
saghul: why is this needed though? we don't have definitions for these AFAICT

app/features/welcome/components/Welcome.js
```diff
@@ -0,0 +1,143 @@
...
+        this._onJoin();
+    }
+
+    _onJoin: (*) => void;
+
+    /**
+     * Redirect and join conference.
+     */
+    _onJoin() {
+
```
saghul: style: extra newline

app/features/app/components/App.js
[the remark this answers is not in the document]
saghul: Actually, why not use flow-typed as we do on jitsi-meet? It's even the recommended way to go: https://www.styled-components.com/docs/api#flow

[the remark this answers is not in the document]
AkshitKrNagpal: Hmm ... We definitely can. But we'll still have to $FlowFixMe for Atlaskit (and maybe other).
This problem is with every module and flow-typed does not support all.

saghul:
This looks awesome @AkshitKrNagpal ! I'm going to merge it ASAP, but I'll be traveling so I may not have time until Tuesday.

PR #34, opened 2018-06-03
Minor Fixes

AkshitKrNagpal:


.eslintrc.js
```diff
@@ -5,9 +5,5 @@ module.exports = {
     ],
     'plugins': [
         'react'
-    ],
-    'rules': {
```
AkshitKrNagpal: These rules are handled by `eslint-config-jitsi`.
We no longer need to overrride them as warnings. We have already disabled them in this commit - 6f35ddd792c5a3e9cc462c31cb1166fa1fbbdb09

webpack.config.js
```diff
@@ -24,10 +24,6 @@ const commonConfig = {
                     ]
                 },
                 test: /\.js$/
-            },
-            {
-                use: 'node-loader',
```
AkshitKrNagpal: Since we are using `jitsi-meet-electron-utils` package as externals. We no longer need `node-loader`.

saghul:
👍

PR #36, opened 2018-06-04
Configured Redux and react-router-redux

AkshitKrNagpal:


package.json
```diff
@@ -35,16 +35,21 @@
...
+    "react-router-redux": "5.0.0-alpha.9",
+    "redux": "4.0.0",
+    "redux-logger": "3.0.6",
     "styled-components": "3.3.0"
   },
   "devDependencies": {
     "babel-core": "6.26.3",
     "babel-eslint": "8.2.3",
     "babel-loader": "7.1.4",
+    "babel-plugin-transform-object-rest-spread": "6.26.0",
```
saghul: Isn't this included in babel-preset-env ?

app/features/app/components/App.js
```diff
@@ -3,7 +3,9 @@
 import { AtlasKitThemeProvider } from '@atlaskit/theme';
 
 import React, { Component } from 'react';
-import { HashRouter as Router, Route, Switch } from 'react-router-dom';
+import { Route, Switch } from 'react-router';
+import { ConnectedRouter as Router } from 'react-router-redux';
```
saghul: unwritten style rule: group these imports and sort them alphabetically.

also, relative imports are sorted from "furthest" to "closest"

app/features/redux/middleware.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+
+import { applyMiddleware } from 'redux';
+
+import { middleware as routerMiddleware } from '../router';
+
+import { createLogger } from 'redux-logger';
```
saghul: ditto

app/features/router/history.js
```diff
@@ -0,0 +1,7 @@
+// @flow
+
+import { createHashHistory } from 'history';
```
saghul: why do we need this hash history thing if we are not using (or are we?) a web based router? I thought the redux router may allow us to not use hash based routing.

app/index.js
```diff
@@ -10,8 +10,17 @@ import { App } from './features/app';
...
+import { Provider } from 'react-redux';
+import { store } from './features/redux';
+
 /**
  * Render the main / root application.
- * $FlowFixMe
  */
-render(<App />, document.getElementById('app'));
+render(
+    <Provider store={store}>
```
saghul: style: space between `=` and around `{`.

app/index.js
```diff
@@ -10,8 +10,17 @@ import { App } from './features/app';
...
+        <App />
+    </Provider>,
```
saghul: Can you create a component in this very file, call it `Root`, which encapsulates this provider and app?

AkshitKrNagpal: Sure.

app/index.js
AkshitKrNagpal: Damn. I always miss this somewhere.
Will update this.

app/features/router/history.js
AkshitKrNagpal: `react-router-redux` is still web based. This is just somehow linking history object (which contains all the information about browser session) to redux store using middleware and reducer.

You can see usage here.
https://github.com/ReactTraining/react-router/tree/master/packages/react-router-redux#usage

package.json
AkshitKrNagpal: I got an error. So I searched and found this. 
https://github.com/babel/babel-preset-env/issues/49

I guess they decide not to include this.

app/features/redux/middleware.js
AkshitKrNagpal: I was actually keeping this as like this.
First, `applyMiddleware()` import is from redux. (single line)
Then middleware from different features. (without line gap)
Then middleware from different packages like `redux-logger` right now. (possibly `redux-thunk` etc. later)

WDYT?

app/features/app/components/App.js
AkshitKrNagpal: Okay, I moved history import wih other relative imports.
So its looks like this now.
```
import { AtlasKitThemeProvider } from '@atlaskit/theme';

import React, { Component } from 'react';
import { Route, Switch } from 'react-router';
import { ConnectedRouter as Router } from 'react-router-redux';

import { Conference } from '../../conference';
import { history } from '../../router';
import { Welcome } from '../../welcome';
```
Questions
- Do you want me to group relative imports with react and router imports too?
- Do we sort module name or component name alphabeltically? ( `react-router` should come above `react-redux-router` or `ConnectedRouter` should come above `Router, Switch`)

package.json
AkshitKrNagpal: I guess we can upgrade too. `@babel/preset-env`

https://github.com/babel/babel/tree/master/packages/babel-preset-env

saghul: Let's go with `babel-preset-stage-1` since we also use it on jitsi-meet. It includes this plugin.

app/features/redux/middleware.js
saghul: Let's be consistent with the other files please!

app/features/router/history.js
saghul: 👍

package.json
AkshitKrNagpal: Okay. I'll update this.

app/features/redux/middleware.js
AkshitKrNagpal: Okay. 
Then this will be fine then.
```
import { applyMiddleware } from 'redux';
import { createLogger } from 'redux-logger';

import { middleware as routerMiddleware } from '../router';
```

AkshitKrNagpal:
Ping @saghul 
Resolved all comments.

saghul:
I'll take a look, thanks!

saghul:
Landed with minor stylistic changes. Nice work @AkshitKrNagpal !

PR #38, opened 2018-06-06
Modified to set theme from redux store

AkshitKrNagpal:
This way, we won't need theme provider in other than root component.

### DIFF
--- a/app/features/app/actionTypes.js
+++ b/app/features/app/actionTypes.js
@@ -1,0 +1,10 @@
+/**
+ * The type of (redux) action that sets the theme of the
+ * AtlaskitThemeProvider component.
+ *
+ * {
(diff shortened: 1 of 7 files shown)


saghul:
Hum, I'm not sure I like this. It would make sense if the user is able to select the theme, but we probably don't want to do that yet.

Can you table this for now and see how it goes once we have the settings panel on the welcome page?

saghul:
> Sure we can revisit this later. If you don't think this should be done until we have user option for the theme.

Yep, that's what I have in mind. Let's keep it open and take a look after the welcome page is more complete so we can give it a shot.

Hey, even if we decide not go this way, I'd like you to know we appreciate your proactivity, keep the suggestions coming!

PR #39, opened 2018-06-06
feat(google-auth): Implement example of setting up google auth

virtuacoplenny:
Related jitsi-meet-electron-utils changes: https://github.com/jitsi/jitsi-meet-electron-utils/pull/23


virtuacoplenny:
You're right; I can add them. I wasn't sure they were needed yet but I might as well do it because this is an example app.

virtuacoplenny:
TODO: increase version number in package.json

saghul:
> because this is an example app.

Not for much longer :-)

PR #40, opened 2018-06-08
Added initial navigation bar

AkshitKrNagpal:


app/features/app/components/App.js
```diff
@@ -35,17 +37,19 @@ export default class App extends Component<*> {
...
-                        <Route
-                            exact
-                            component = { Welcome }
-                            path = '/' />
-                        <Route
-                            component = { Conference }
-                            path = '/:domain?/:room' />
-                    </Switch>
-                </Router>
+                <Page navigation = { <Navbar /> }>
```
saghul: We only want the navigation in the welcome page, so please move it there. While in a conference, we want it to take over the entire screen.

app/features/navbar/components/HelpAction.js
```diff
@@ -0,0 +1,89 @@
...
+    render() {
+        return (
+            <Droplist
+                isOpen = { this.state.droplistOpen }
+                onClick = { () => this._droplistToggle() }
+                onOpenChange = { () => this._droplistToggle() }
+                position = 'right bottom'
+                trigger = { <HelpIcon /> } >
+                <Item onActivate = { this._openTermsPage }>Terms</Item>
+                <Item onActivate = { this._openPrivacyPage }>Privacy</Item>
```
saghul: can we please add a support link here? it should point to `mailto:support@jitsi.org`

AkshitKrNagpal: Yeah. Sure.

app/features/app/components/App.js
AkshitKrNagpal: I don't think we should do that.
Wouldn't user like to have access to drawers we will create in navigation while in the conference?

app/features/navbar/components/HelpAction.js
AkshitKrNagpal: As `support` or `feedback`?

saghul: Right. Ler's go with "Send Feedback" since we have that on mobile too.

app/features/app/components/App.js
saghul: They are not really useful IMHO. The conference itself has the ability to change things, that's why we discussed the syncing of our local settings with the ones the conference provides using the iframe API.

Maybe i'm used to seeing the conference component take the entire screen, so this looks weird to me.

@hristoterezov @gpolitis WDYT?

saghul:
screen shot 2018-06-08 at 13 52 24
screen shot 2018-06-08 at 13 51 47

For the discussion about whether to use the navbar in the conference: here is how it looks like.

app/features/navbar/components/HelpAction.js
AkshitKrNagpal: Done.

app/features/app/components/App.js
AkshitKrNagpal: @saghul 
Hmm. But then there will be different UI for settings on both components. 

PS - Syncing is not a problem. It will be easy.

saghul: > Hmm. But then there will be different UI for settings on both components.

This will be the case no matter what, since the Meet UI does have the ability to set those.

app/features/config/index.js
```diff
@@ -8,5 +8,20 @@ export default {
...
+
+    /**
+     * The url of Terms and Conditions Page.
+     */
+    termsAndConditionsURL: 'https://jitsi.org/meet/terms',
+
+    /**
+     * Email to send feedback.
+     */
+    feedbackEmail: 'support@jitsi.org'
```
saghul: Let's make this "feedbackURL" and add the scheme. This way, if we ever want to change from a mailto URL to a HTTP URL it will be a single line change.

AkshitKrNagpal: Hmm. 👍

app/features/app/components/App.js
AkshitKrNagpal: @saghul 
Yeah. 
Should I update this or wait for @hristoterezov and @gpolitis?

saghul: I'd like them to weigh in and break the tie here, since you and I stand on opposite sides at the moment :-)

app/features/navbar/components/HelpAction.js
```diff
@@ -0,0 +1,99 @@
...
+        openExternalLink(config.privacyPolicyURL);
+    }
+
+    _sendFeedback: (*) => void;
+
+    /**
+     * Opens Support/Feedback Email.
+     */
+    _sendFeedback() {
+        openExternalLink(config.feedbackURL);
```
saghul: Missing the binding in the constructor. I know it's not technically needed now, but it will once we move config to redux and we need access to this.props here.

AkshitKrNagpal: Oh Yeah. My mistake.

AkshitKrNagpal: Updated.

saghul:
@AkshitKrNagpal LGTM, modulo the navbar location thing. Once we sort that out this should be good to go.

app/features/app/components/App.js
gpolitis: I'm not a UI expert but I agree with Saul that jitsi-meet should provide all the necessary drawers/menus to configure the conference and also, in general, I prefer the interface that I'm working with to be as uncluttered as possible.

AkshitKrNagpal: @gpolitis Cool.
@saghul I have updated this.

saghul:
Landed! :tada: @AkshitKrNagpal onwards and upwards! :muscle:

PR #41, opened 2018-06-09
Added settings drawer

AkshitKrNagpal:
TODO
- [x] Sync name and email with conference
- [x] Redux Persist
- [x] UI Improvement
- [x] Avatar

### DIFF
--- a/app/features/conference/components/Conference.js
+++ b/app/features/conference/components/Conference.js
@@ -14,10 +14,16 @@ import {
 } from 'jitsi-meet-electron-utils';
 
 import config from '../../config';
+import { setEmail, setName } from '../../settings';
 
 import { Wrapper } from '../styled';
 
 type Props = {
+
+    /**
+     * Redux dispatch.
+     */
+    dispatch: Dispatch<*>;
 
     /**
      * React Router match object.
@@ -26,9 +32,20 @@ type Props = {
     match: Object;
 
     /**
-     * Redux dispatch.
+     * Avatar URL.
      */
-    dispatch: Dispatch<*>;
+    _avatarURL: string;
+
+    /**
+     * Email of user.
+     */
+    _email: string;
+
+    /**
+     * Name of user.
+     */
+    _name: string;
+
 };
 
 /**
@@ -74,6 +91,26 @@ class Conference extends Component<Props, *> {
         script.src = `https://${domain}/external_api.js`;
 
         this._ref.current.appendChild(script);
+    }
+
+    /**
+     * Keep profile settings in sync with Conference.
+     *
+     * @param {Props} prevProps - Component's prop values before update.
+     * @returns {void}
+     */
+    componentDidUpdate(prevProps) {
+        const { props } = this;
+
+        if (props._avatarURL !== prevProps._avatarURL) {
+            this._setAvatarURL(props._avatarURL);
+        }
+        if (props._email !== prevProps._email) {
+            this._setEmail(props._email);
+        }
+        if (props._name !== prevProps._name) {
+            this._setName(props._name);
+        }
     }
 
     /**
@@ -131,7 +168,107 @@ class Conference extends Component<Props, *> {
         setupWiFiStats(iframe);
 
         this._api.on('readyToClose', () => this._navigateToHome());
+
+        this._api.on('videoConferenceJoined',
+            (conferenceInfo: Object) =>
+                this._onVideoConferenceJoined(conferenceInfo));
     }
+
+    /**
+     * Updates redux state's user name from conference.
+     *
+     * @param {Object} params - Returned object from event.
+     * @param {string} id - Local Participant ID.
+     * @returns {void}
+     */
+    _onDisplayNameChange(params: Object, id: string) {
+        if (params.id === id) {
+            this.props.dispatch(setName(params.displayname));
+        }
+    }
+
+    /**
+     * Updates redux state's email from conference.
+     *
+     * @param {Object} params - Returned object from event.
+     * @param {string} id - Local Participant ID.
+     * @returns {void}
+     */
+    _onEmailChange(params: Object, id: string) {
+        if (params.id === id) {
+            this.props.dispatch(setEmail(params.email));
+        }
+    }
+
+    /**
+     * Saves conference info on joining it.
+     *
+     * @param {Object} conferenceInfo - Contains information about the current
+     * conference.
+     * @returns {void}
+     */
+    _onVideoConferenceJoined(conferenceInfo: Object) {
+
+        this._setAvatarURL(this.props._avatarURL);
+        this._setEmail(this.props._email);
+        this._setName(this.props._name);
+
+        const { id } = conferenceInfo;
+
+        this._api.on('displayNameChange',
+            (params: Object) => this._onDisplayNameChange(params, id));
+        this._api.on('emailChange',
+            (params: Object) => this._onEmailChange(params, id));
+    }
+
+    /**
+     * Set Avatar URL from settings to conference.
+     *
+     * @param {string} avatarURL - Avatar URL.
+     * @returns {void}
+     */
+    _setAvatarURL(avatarURL: string) {
+        this._api.executeCommand('avatarUrl', avatarURL);
+    }
+
+    /**
+     * Set email from settings to conference.
+     *
+     * @param {string} email - Email of user.
+     * @returns {void}
+     */
+    _setEmail(email: string) {
+        this._api.executeCommand('email', email);
+    }
+
+    /**
+     * Set name from settings to conference.
+     *
+     * @param {string} name - Name of user.
+     * @returns {void}
+     */
+    _setName(name: string) {
+        this._api.executeCommand('displayName', name);
+    }
+
 }
 
-export default connect()(Conference);
+/**
+ * Maps (parts of) the redux state to the React props.
+ *
+ * @param {Object} state - The redux state.
+ * @returns {{
+ *     _avatarURL: string,
+ *     _email: string,
+ *     _name: string
+ * }}
+ */
+function _mapStateToProps(state: Object) {
+    return {
+        _avatarURL: state.settings.avatarURL,
+        _email: state.settings.email,
+        _name: state.settings.name
+    };
+}
+
+export default connect(_mapStateToProps)(Conference);
--- a/app/features/navbar/actionTypes.js
+++ b/app/features/navbar/actionTypes.js
@@ -1,0 +1,18 @@
+/**
+ * The type of (redux) action that opens specified Drawer.
+ *
+ * {
+ *     type: OPEN_DRAWER,
+ *     drawerComponent: React.ComponentType<*>
+ * }
+ */
+export const OPEN_DRAWER = Symbol('OPEN_DRAWER');
+
+/**
+ * The type of (redux) action that closes all Drawer.
+ *
+ * {
+ *     type: CLOSE_DRAWER
+ * }
+ */
+export const CLOSE_DRAWER = Symbol('CLOSE_DRAWER');
--- a/app/features/navbar/actions.js
+++ b/app/features/navbar/actions.js
@@ -1,0 +1,34 @@
+// @flow
+
+import type { ComponentType } from 'react';
+
+import { CLOSE_DRAWER, OPEN_DRAWER } from './actionTypes';
+
+/**
+ * Closes the drawers.
+ *
+ * @returns {{
+ *     type: CLOSE_DRAWER,
+ * }}
+ */
+export function closeDrawer() {
+    return {
+        type: CLOSE_DRAWER
+    };
+}
+
+/**
+ * Opens the specified drawer.
+ *
+ * @param {string} drawerComponent - Component of the drawer.
+ * @returns {{
+ *     type: OPEN_DRAWER,
+ *     drawerComponent: ComponentType<*>
+ * }}
+ */
+export function openDrawer(drawerComponent: ComponentType<*>) {
+    return {
+        type: OPEN_DRAWER,
+        drawerComponent
+    };
+}
--- a/app/features/navbar/reducer.js
+++ b/app/features/navbar/reducer.js
@@ -1,0 +1,39 @@
+// @flow
+
+import type { ComponentType } from 'react';
+
+import { CLOSE_DRAWER, OPEN_DRAWER } from './actionTypes';
+
+type State = {
+    openDrawer: typeof undefined | ComponentType<*>
+};
+
+const DEFAULT_STATE = {
+    openDrawer: undefined
+};
+
+/**
+ * Reduces redux actions for features/settings.
+ *
+ * @param {State} state - Current reduced redux state.
+ * @param {Object} action - Action which was dispatched.
+ * @returns {State} - Updated reduced redux state.
+ */
+export default (state: State = DEFAULT_STATE, action: Object) => {
+    switch (action.type) {
+    case CLOSE_DRAWER:
+        return {
+            ...state,
+            openDrawer: undefined
+        };
+
+    case OPEN_DRAWER:
+        return {
+            ...state,
+            openDrawer: action.drawerComponent
+        };
+
+    default:
+        return state;
+    }
+};
--- a/app/features/redux/index.js
+++ b/app/features/redux/index.js
@@ -1,1 +1,2 @@
+export { default as persistor } from './persistor';
 export { default as store } from './store';
--- a/app/features/redux/reducers.js
+++ b/app/features/redux/reducers.js
@@ -2,8 +2,12 @@
 
 import { combineReducers } from 'redux';
 
+import { reducer as navbarReducer } from '../navbar';
 import { reducer as routerReducer } from '../router';
+import { reducer as settingsReducer } from '../settings';
 
 export default combineReducers({
-    router: routerReducer
+    navbar: navbarReducer,
+    router: routerReducer,
+    settings: settingsReducer
 });
--- a/app/features/settings/actionTypes.js
+++ b/app/features/settings/actionTypes.js
@@ -1,0 +1,30 @@
+/**
+ * The type of (redux) action that sets the Avatar URL.
+ *
+ * {
+ *     type: SET_AVATAR_URL,
+ *     avatarURL: string
+ * }
+ */
+export const SET_AVATAR_URL = Symbol('SET_AVATAR_URL');
+
+/**
+ * The type of (redux) action that sets the email of the user.
+ *
+ * {
+ *     type: SET_EMAIL,
+ *     email: string
+ * }
+ */
+export const SET_EMAIL = Symbol('SET_EMAIL');
+
+/**
+ * The type of (redux) action that sets the name of the user.
+ *
+ * {
+ *     type: SET_NAME,
+ *     name: string
+ * }
+ */
+export const SET_NAME = Symbol('SET_NAME');
+
--- a/app/features/settings/actions.js
+++ b/app/features/settings/actions.js
@@ -1,0 +1,51 @@
+// @flow
+
+import { SET_AVATAR_URL, SET_EMAIL, SET_NAME } from './actionTypes';
+
+/**
+ * Set Avatar URL.
+ *
+ * @param {string} avatarURL - Avatar URL.
+ * @returns {{
+ *     type: SET_AVATAR_URL,
+ *     avatarURL: string
+ * }}
+ */
+export function setAvatarURL(avatarURL: string) {
+    return {
+        type: SET_AVATAR_URL,
+        avatarURL
+    };
+}
+
+/**
+ * Set the email of the user.
+ *
+ * @param {string} email - Email of the user.
+ * @returns {{
+ *     type: SET_EMAIL,
+ *     email: string
+ * }}
+ */
+export function setEmail(email: string) {
+    return {
+        type: SET_EMAIL,
+        email
+    };
+}
+
+/**
+ * Set the name of the user.
+ *
+ * @param {string} name - Name of the user.
+ * @returns {{
+ *     type: SET_NAME,
+ *     name: string
+ * }}
+ */
+export function setName(name: string) {
+    return {
+        type: SET_NAME,
+        name
+    };
+}
--- a/app/features/settings/components/SettingsDrawer.js
+++ b/app/features/settings/components/SettingsDrawer.js
@@ -1,0 +1,193 @@
+// @flow
+
+import Avatar from '@atlaskit/avatar';
+import FieldText from '@atlaskit/field-text';
+import ArrowLeft from '@atlaskit/icon/glyph/arrow-left';
+import { AkCustomDrawer } from '@atlaskit/navigation';
+
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+import type { Dispatch } from 'redux';
+
+import { closeDrawer, DrawerContainer, Logo } from '../../navbar';
+import { AvatarContainer, ProfileContainer } from '../styled';
+import { setEmail, setName } from '../actions';
+
+type Props = {
+
+    /**
+     * Redux dispatch.
+     */
+    dispatch: Dispatch<*>;
+
+    /**
+     * Is the drawer open or not.
+     */
+    isOpen: boolean;
+
+    /**
+     * Avatar URL.
+     */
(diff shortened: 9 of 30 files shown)


AkshitKrNagpal:
Ping @saghul 
PTAL at how I plan on structuring drawers.

saghul:
This looks good @AkshitKrNagpal ! One thing I'd like you to explore is using the component type itself as the parameter to the OPEN_DRAWER action, instead of a name which describes it. You can see something similar in the dialog feature on jitsi-met. This avoids having to map a name to a component type, for example.

AkshitKrNagpal:
@saghul 
Thanks for the suggestion. I was able to do this thanks to you. 😄 
I feel the code also looks cleaner this way.

saghul:
@AkshitKrNagpal excellent, this indeed looks much better! Let's continue with these items in this PR:

- avatar
- display name
- email

As we discussed the computation of the avatar URL needs to be extracted from jitsi-meet and moved to js-utils.

Oh, and one thing I noticed is that when the drawer is closed the focus remains on the tiny avatar in the navbar. It would ne nice it we could make it be the room field.

AkshitKrNagpal:
> avatar
display name
email
As we discussed the computation of the avatar URL needs to be extracted from jitsi-meet and moved to js-utils.

Yeah. Already working on that.

> Oh, and one thing I noticed is that when the drawer is closed the focus remains on the tiny avatar in the navbar. It would ne nice it we could make it be the room field.

I'll look into it.

AkshitKrNagpal:
@saghul 
I am not able to reproduce this. The avatar is not focused when I close the drawer. 

About focusing on room field when closing drawer, I don't know how would be able to do that. 
Field text of Atlaskit has an option to set automatic focus on field text (which I already am doing) but that only focuses when the component is rendered.

saghul:
@AkshitKrNagpal thanks for checking! I'll take another look.

saghul:
I had another quick look and yeah, I no longer see the avatar having focus. The text field doesn't have it either, but can think about that later, it's not critical.

Here are some observations, in random order:

- let's call this "settings" instead of "profile". in jitsi-meet we also had a profile and renamed everything to settings, sorry I didn't catch this earlier...
- since we are here, it's probably a good idea to add the start with audio / video muted switches to the drawer
- we want the settings persisted across runs, I think we could use https://github.com/rt2zz/redux-persist

saghul:
Ah, also, let's add the avatar to the drawer as well, so the user can see it change when they enter the email address.

Last, can the drawer be made narrower and the text fields a bit longer? As it stands they look a bit too small. (I know this is WIP, so no biggie)

AkshitKrNagpal:
> let's call this "settings" instead of "profile". in jitsi-meet we also had a profile and renamed everything to settings, sorry I didn't catch this earlier...
since we are here, it's probably a good idea to add the start with audio / video muted switches to the drawer

Cool. I'll do this too.

> we want the settings persisted across runs, I think we could use https://github.com/rt2zz/redux-persist

Already planned on doing this. 😄 

> Ah, also, let's add the avatar to the drawer as well, so the user can see it change when they enter the email address. 

Sure.

AkshitKrNagpal:
> Last, can the drawer be made narrower and the text fields a bit longer? As it stands they look a bit too small. (I know this is WIP, so no biggie)

Hmm. But on large screen it has to be this size. I think we'll need to make this kinda responsive.
Narrow on small screens and like this on large screens. (Narrow will look really thin on large screens)

AkshitKrNagpal:
TODO
- [ ] Redux Persist
- [ ] UI Improvement
- [ ] Avatar
- [ ] Responsive Navbar Width

AkshitKrNagpal:
@saghul 
> Let's call this "settings" instead of "profile". in jitsi-meet we also had a profile and renamed everything to settings, sorry I didn't catch this earlier...

I think we should keep profile and conference settings separate.

When a user will click on avatar, he will expect the things that has to do with his "profile" not settings like default conference server domain or start with audio / video muted switches.

On mobile app we have settings in which there are two parts. One is profile (containing name and email) and one is conference( containing server url and audio / video muted switches). But the user doesn't click on something like an avatar. It is a hamburger menu button and then "Settings". 

WDYT?

saghul:
@AkshitKrNagpal Even though on mobile we say "Profile" in a header, the feature is called settings. Because that's what they really are, if you think about it: settings which apply to different parts of the experience.

So, I think a features/settings would be more appropriate for consistency and less surprise for other developers.

Now, having the code there doesn't mean we can't have a "Profile" header before the name and email fields, we can certainly experiment with that!

AkshitKrNagpal:
@saghul 
Yeah. I have no problem with this.
I am saying wouldn't user feel weird that an avatar button takes him to settings?

saghul:
@AkshitKrNagpal Ah, that! Yeah, I forgot to mention it, indeed! Let's add a button which suggests "settings". The one Stride uses will do for example.

AkshitKrNagpal:
@saghul 
So replace avatar with settings icon?
**It will end up looking more and more like stride.** 

I say we go with "profile" drawer in which only 3 things will be possible for now. Name, email and avatar change. ( I am not sure what that latest google-auth thing is right now but maybe it can go there)

And create a separate "settings" drawer for the default domain. (and other settings like whether we want to disable always on top, you know something like that we can implement in future). Because if you think about it these are the settings of the desktop application.

And start with audio/video muted switches are better suited in front of the eyes where we will type room name (maybe something like two icons something like we have at bottom centre during the conference).

AkshitKrNagpal:
> No. Leave the avatar where it is now. We might replace it with a Jitsi logo, for example. Then add a settings button right below it, not at the bottom, where the help button is. Inside the drawer lets put a larger avatar.

Okay. I don't see a problem in that. But guess I'll remove the avatar then. Whats say?

> No, I don't like this. let's have 1 drawer with all the app settings. There is no profile and no conference settings really, that's just an illusion, in that sense we could add a section for each!

Hmm. I do understand that you want to keep drawers as minimum. Okay. Lets do with only one drawer. 😄 

> The google auth thing is for live streaming, but it's seamless, there is nothing to have a setting for

Oh. I thought maybe we are doing something like google auth connecting for email.

> Also I don't see the need for a toggle for AoT window.

Well. I do think user should have control over it. Its not really a need but a want maybe. Let's discuss about some of these features after then. 😃

saghul:
> Okay. I don't see a problem in that. But guess I'll remove the avatar then. Whats say?

Sounds good!

> Well. I do think user should have control over it. Its not really a need but a want maybe. Let's discuss about some of these features after then. 😃

Agreed, let's keep the discussion open, but let's start small so we don't entangle ourselves :-)

AkshitKrNagpal:
PTAL. @saghul

app/features/conference/components/Conference.js
```diff
@@ -29,6 +30,21 @@ type Props = {
...
+
+    /**
+     * Email of user.
+     */
+    _email: string;
+
+    /**
+     * Avatar URL.
+     */
+    _avatarURL: string;
```
saghul: style: alphabetical sort

app/features/conference/components/Conference.js
```diff
@@ -131,7 +147,104 @@ class Conference extends Component<Props, *> {
         setupWiFiStats(iframe);
 
         this._api.on('readyToClose', () => this._navigateToHome());
+
+        this._setName(this.props._name);
```
saghul: ditto

app/features/conference/components/Conference.js
```diff
@@ -131,7 +147,104 @@ class Conference extends Component<Props, *> {
...
+        this._setEmail(this.props._email);
+        this._setAvatarURL(this.props._avatarURL);
```
saghul: These may need to be set also in `componentWillReceiveProps`, in order to set the new avatar URL calculated after the email changed within the conference.

app/features/conference/components/Conference.js
```diff
@@ -131,7 +147,104 @@ class Conference extends Component<Props, *> {
...
+ *
+ * @param {Object} state - The redux state.
+ * @returns {{
+ *     _name: string,
+ *     _email: string
+ * }}
+ */
+function _mapStateToProps(state: Object) {
+    return {
+        _name: state.settings.name,
```
saghul: style: alpha lort

app/features/conference/components/Conference.js
```diff
@@ -131,7 +147,104 @@ class Conference extends Component<Props, *> {
...
+    }
+}
+
+/**
+ * Maps (parts of) the redux state to the React props.
+ *
+ * @param {Object} state - The redux state.
+ * @returns {{
+ *     _name: string,
+ *     _email: string
```
saghul: avatarURL is missing

app/features/navbar/styled/drawerContainer.js
```diff
@@ -0,0 +1,8 @@
+// @flow
```
saghul: This file should be called DrawerContainer. I guess you didn't notice the problem because macOS uses a case-insensitive filesystem.

app/features/navbar/styled/drawerContainer.js
```diff
@@ -0,0 +1,8 @@
...
+
+import styled from 'styled-components';
+
+export default styled.div`
+    padding: 0 8px;
```
saghul: style: alpha sort

app/features/redux/index.js
```diff
@@ -1 +1,2 @@
 export { default as store } from './store';
+export { default as persistor } from './persistor';
```
saghul: style: alpha sort

app/features/redux/reducers.js
```diff
@@ -2,8 +2,12 @@
 
 import { combineReducers } from 'redux';
 
+import { reducer as settingsReducer } from '../settings';
+import { reducer as navbarReducer } from '../navbar';
```
saghul: style: alpha sort

app/features/settings/actionTypes.js
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * The type of (redux) action that sets the Avatar URL.
+ *
+ * {
+ *     type: SET_AVATAR_URL,
+ *     url: string
+ * }
+ */
+export const SET_AVATAR_URL = Symbol('SET_AVATAR_URL');
```
saghul: style: alpha sort

app/features/settings/actions.js
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Set Avatar URL.
+ *
+ * @param {string} avatarURL - Avatar URL.
+ * @returns {{
+ *     type: SET_AVATAR_URL,
+ *     avatarURL: string
+ * }}
+ */
+export function setAvatarURL(avatarURL: string) {
```
saghul: style: alpha sort

app/features/settings/components/SettingsDrawer.js
```diff
@@ -0,0 +1,168 @@
...
+import { AkCustomDrawer } from '@atlaskit/navigation';
+
+import React, { Component } from 'react';
+import { connect } from 'react-redux';
+import type { Dispatch } from 'redux';
+
+import { closeDrawer, DrawerContainer } from '../../navbar';
+import { getAvatarURL } from '../../utils';
+import { ProfileContainer, AvatarContainer } from '../styled';
+import { setName, setEmail, setAvatarURL } from '../actions';
```
saghul: style: alpha sort

app/features/settings/components/SettingsDrawer.js
```diff
@@ -0,0 +1,168 @@
...
+
+type Props = {
+
+    /**
+     * Is the drawer open or not.
+     */
+    isOpen: boolean;
```
saghul: style: alpha sort

app/features/settings/components/SettingsDrawer.js
```diff
@@ -0,0 +1,168 @@
...
+ * @returns {{
+ *     _name: string,
+ *     _email: string
+ * }}
+ */
+function _mapStateToProps(state: Object) {
+    return {
+        _name: state.settings.name,
+        _email: state.settings.email,
+        _avatarURL: state.settings.avatarURL
```
saghul: style: alpha sort

app/features/settings/reducer.js
```diff
@@ -0,0 +1,47 @@
...
+            name: action.name
+        };
+
+    case SET_EMAIL:
+        return {
+            ...state,
+            email: action.email
+        };
+
+    case SET_AVATAR_URL:
```
saghul: style: alpha sort

app/features/settings/styled/AvatarContainer.js
```diff
@@ -0,0 +1,9 @@
+// @flow
+
+import styled from 'styled-components';
+
+export default styled.div`
+    display: flex;
+    flex-direction: column;
+    align-items: center;
```
saghul: style: alpha sort

app/features/settings/components/SettingsDrawer.js
```diff
@@ -0,0 +1,168 @@
...
+
+    _onEmailBlur: (*) => void;
+
+    /**
+     * Updates Avatar URL in (redux) state when email is updated.
+     *
+     * @returns {void}
+     */
+    _onEmailBlur() {
+        const avatarURL = getAvatarURL(this.props._email);
```
saghul: AFAICT we wouldn't be setting any URL if the user enters no email. The way we solve this on Jitsi Meet is by having some locally generated UUID which is used as the key to this function.

app/features/utils/functions.js
```diff
@@ -22,3 +23,31 @@ export function openExternalLink(link: string) {
...
     return process.platform === 'darwin';
 }
+
+/**
+ * Returns the Avatar URL to be used.
+ *
+ * @param {string} key - Unique key to generate Avatar URL.
+ * @returns {string}
+ */
+export function getAvatarURL(key: string) {
```
saghul: We talked about having this function shared with jitsi-meet, by placing it in js-utils and using it on both places.

AkshitKrNagpal: Yes. I'll modify this when everything else is in place.

app/features/conference/components/Conference.js
AkshitKrNagpal: Hmm. Avatar URL might need this.
Name and Email won't need it. I think we should hold this PR till we implement `emailChange` event so that we can debug code we will create with `componentWillReceiveProps`.
Otherwise, we might have to create another PR.

app/features/conference/components/Conference.js
AkshitKrNagpal: Done.

app/features/navbar/styled/drawerContainer.js
AkshitKrNagpal: Done.

app/features/settings/actionTypes.js
AkshitKrNagpal: Done.

app/features/settings/actions.js
AkshitKrNagpal: Done.

app/features/settings/components/SettingsDrawer.js
AkshitKrNagpal: Done.

app/features/conference/components/Conference.js
saghul: Creating another PR is fine. Personally, I consider changing the email not to be a common use case. Moreover, even if we fix it today, it needs to be deployed, which won't happen immediately. So, let's write the code as correct was we can with these assumptions and once we fix the email thing we can check if this needs further fixing.

app/features/navbar/styled/drawerContainer.js
saghul: Did you force push? I don't see the change.

app/features/utils/functions.js
saghul: :+1:

app/features/navbar/styled/drawerContainer.js
AkshitKrNagpal: Weird.
Git doesn't even consider this change in macOS. 😨
I updated it using Github website and squash generated commit.

app/features/conference/components/Conference.js
```diff
@@ -131,7 +149,110 @@ class Conference extends Component<Props, *> {
...
     }
+
+    /**
+     * Saves conference info on joining it.
+     *
+     * @param {Object} conferenceInfo - Contains information about the current
+     * conference.
+     * @returns {void}
+     */
+    _onVideoConferenceJoined(conferenceInfo: Object) {
```
saghul: style: alpha sort

app/features/conference/components/Conference.js
```diff
@@ -131,7 +149,110 @@ class Conference extends Component<Props, *> {
...
+        this._api.executeCommand('email', email);
+    }
+
+    /**
+     * Set Avatar URL from settings to conference.
+     *
+     * @param {string} avatarURL - Avatar URL.
+     * @returns {void}
+     */
+    _setAvatarURL(avatarURL: string) {
```
saghul: style: alpha sort

app/features/navbar/actionTypes.js
```diff
@@ -0,0 +1,18 @@
...
+export const OPEN_DRAWER = Symbol('OPEN_DRAWER');
+
+/**
+ * The type of (redux) action that closes all Drawer.
+ *
+ * {
+ *     type: CLOSE_DRAWER
+ * }
+ */
+export const CLOSE_DRAWER = Symbol('CLOSE_DRAWER');
```
saghul: style: alpha sort

app/features/navbar/actions.js
```diff
@@ -0,0 +1,34 @@
+// @flow
+
+import type { ComponentType } from 'react';
+
+import { OPEN_DRAWER, CLOSE_DRAWER } from './actionTypes';
```
saghul: style: alpha sort

app/features/navbar/actions.js
```diff
@@ -0,0 +1,34 @@
...
+}
+
+/**
+ * Closes the drawers.
+ *
+ * @returns {{
+ *     type: OPEN_DRAWER,
+ * }}
+ */
+export function closeDrawer() {
```
saghul: style: alpha sort

app/features/navbar/reducer.js
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Reduces redux actions for features/settings.
+ *
+ * @param {State} state - Current reduced redux state.
+ * @param {Object} action - Action which was dispatched.
+ * @returns {State} - Updated reduced redux state.
+ */
+export default (state: State = DEFAULT_STATE, action: Object) => {
+    switch (action.type) {
+    case OPEN_DRAWER:
```
saghul: style: alpha sort

app/features/redux/reducers.js
```diff
@@ -2,8 +2,12 @@
 
 import { combineReducers } from 'redux';
 
+import { reducer as navbarReducer } from '../navbar';
+import { reducer as settingsReducer } from '../settings';
```
saghul: style: alpha sort

AkshitKrNagpal: Whats wrong in this?

app/features/navbar/actionTypes.js
AkshitKrNagpal: Done.

app/features/conference/components/Conference.js
AkshitKrNagpal: Done.

app/features/conference/components/Conference.js
AkshitKrNagpal: Done.

app/features/conference/components/Conference.js
```diff
@@ -131,7 +149,115 @@ class Conference extends Component<Props, *> {
...
+    }
+
+    /**
+     * Updates redux state's user name from conference.
+     *
+     * @param {Object} params - Returned object from event.
+     * @param {string} id - Local Participant ID.
+     * @returns {void}
+     */
+    _onDisplayNameChange(params: Object, id: string) {
```
saghul: This should also trigger an update on the avatar.

app/features/conference/components/Conference.js
```diff
@@ -131,7 +149,115 @@ class Conference extends Component<Props, *> {
...
+     * @param {Object} conferenceInfo - Contains information about the current
+     * conference.
+     * @returns {void}
+     */
+    _onVideoConferenceJoined(conferenceInfo: Object) {
+        const { id } = conferenceInfo;
+
+        this._api.on('emailChange',
+            (params: Object) => this._onEmailChange(params, id));
+        this._api.on('displayNameChange',
```
saghul: style: alpha sort

app/features/utils/functions.js
```diff
@@ -22,3 +23,35 @@ export function openExternalLink(link: string) {
...
+        urlSuffix = '?d=wavatar&size=200';
+    } else {
+        key = id;
+
+        // Otherwise, use a default (meeples, of course).
+        urlPrefix = 'https://abotars.jitsi.net/meeple/';
+        urlSuffix = '';
+    }
+
+    return urlPrefix + md5.hex(key.trim().toLowerCase()) + urlSuffix;
```
saghul: I know you didn't write this, but can you please use a template string here? thanks!

app/features/conference/components/Conference.js
AkshitKrNagpal: Agreed.
I'll cover this with middleware thing.

AkshitKrNagpal:
> componentWillReceiveProps will be required so the conference updates the avatar URL when it changes mid-conference.

Please see this commit https://github.com/jitsi/jitsi-meet-electron/pull/41/commits/ab2a8a9cfd6590f39f23f37565706f39d08744da . I don't think we'll need to use `componentWillReciveProps` if we do it like this. 
But I wonder will this be possible when we do it using middleware. If not, I'll do it in `componentWillRecieveProps` :smile:

app/features/utils/functions.js
AkshitKrNagpal: I don't understand.
We are just concatenating variables. Not one of them is a string literal (we don't even have space in between) 
Do you want it to be like `${urlPrefix}${md5.hex(key.trim().toLowerCase())}${urlSuffix}` ?

saghul: Yep :-)

saghul:
> Please see this commit ab2a8a9 . I don't think we'll need to use componentWillReciveProps if we do it like this.
> But I wonder will this be possible when we do it using middleware. If not, I'll do it in componentWillRecieveProps 😄

You will need the middleware because the middleware itself will compute the avatar URL, so you won't know it at that point in the code :-)

app/features/utils/functions.js
AkshitKrNagpal: But why?
The template literals were introduced to make string concatenation code easier to read.
If I know correctly, that is only useful when concatenation has string literals and variables. Since we have no strings to concat just variables, this looks better.

AkshitKrNagpal:
> You will need the middleware because the middleware itself will compute the avatar URL, so you won't know it at that point in the code :-)

Please see my message of stride.

app/features/settings/components/SettingsDrawer.js
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    /**
+     * Render function of component.
+     *
+     * @returns {ReactElement}
+     */
+    render() {
+        return (
+            <AkCustomDrawer
```
saghul: Please add the primary icon here too.

app/features/settings/components/SettingsDrawer.js
```diff
@@ -0,0 +1,154 @@
...
+                isOpen = { this.props.isOpen }
+                onBackButton = { this._onBackButton } >
+                <DrawerContainer>
+                    <ProfileContainer>
+                        <AvatarContainer>
+                            <Avatar
+                                size = 'xlarge'
+                                src = { this.props._avatarURL } />
+                        </AvatarContainer>
+                        <FieldText
```
saghul: Enter key still doesn't work.

AkshitKrNagpal: Done

app/features/settings/components/SettingsDrawer.js
AkshitKrNagpal: Done

PR #42, opened 2018-06-10
Updated eslint-config-jitsi

AkshitKrNagpal:


package.json
```diff
@@ -61,7 +61,7 @@
     "electron-packager": "12.0.2",
     "electron-rebuild": "1.7.3",
     "eslint": "4.12.1",
-    "eslint-config-jitsi": "jitsi/eslint-config-jitsi",
+    "eslint-config-jitsi": "jitsi/eslint-config-jitsi#v0.2",
```
hristoterezov: @AkshitKrNagpal It should be #v0.1.0!  I added one more commit with updated version in the package.json and realized that it's better to follow the version from there. I created a new release - v0.1.0.

AkshitKrNagpal:
@hristoterezov 
Oh. I didn't realise that.
Updated.

PR #44, opened 2018-06-16
Remove window menu

saghul:
Looks a lot cleaner on Linux / Windows systems:

screenshot from 2018-06-16 08-40-30


AkshitKrNagpal:
Why are we hiding menu? 

>Looks a lot cleaner on Linux / Windows systems:

It already looks like this on Ubuntu.
Can u share a pic before this change? Which distro are you using?

package.json
```diff
@@ -58,6 +58,7 @@
     "babel-preset-stage-1": "6.24.1",
     "css-loader": "0.28.11",
     "electron": "2.0.0",
+    "electron-debug": "2.0.0",
```
AkshitKrNagpal: Since we are doing this, lets install `devtron` and `electron-react-devtools` too.

`npm install --save-dev devtron electron-react-devtools`

saghul:
> It already looks like this on Ubuntu.
> Can u share a pic before this change? Which distro are you using?

That's only the case on Unity on Ubuntu. Any other desktop environment which does not hijack the menu bar and put it macOS style will display like so:

screenshot from 2018-06-16 11-43-15

So, by hiding it we get consistent looks, and the menu is pretty useless anyway, IMHO.

package.json
saghul: I'll look into those, thanks!

AkshitKrNagpal:
Hmm. Okay. 
But then lets use `Menu.setApplicationMenu(menu)` to remove menu from mac os also. For consistency.

AkshitKrNagpal:
LGTM. 👍

package.json
```diff
@@ -57,11 +57,14 @@
     "babel-preset-react": "6.24.1",
     "babel-preset-stage-1": "6.24.1",
     "css-loader": "0.28.11",
+    "devtron": "1.4.0",
     "electron": "2.0.0",
+    "electron-debug": "2.0.0",
```
AkshitKrNagpal: It should be a dependency, not devDependency.
Open the packaged app to see the error.

saghul: > It should be a dependency, not devDependency.
> Open the packaged app to see the error.

:man_facepalming: fixed, thanks for catching that!

saghul:
@AkshitKrNagpal tested both dev and prod this time... should work, thanks for your review!

PR #45, opened 2018-06-16
[RFC] Added Jitsi Logo to navbar

AkshitKrNagpal:
Ping @saghul 
I have minified the svg using SVGO as specified in [docs](http://atlaskit.atlassian.com/packages/core/icon/docs/using-custom-icons) ) and created the icon here.
I am a little unsure about directory structure. PTAL.


app/features/icons/components/JitsiGlyph.js
```diff
@@ -0,0 +1,30 @@
...
+class JitsiGlyph extends Component<*> {
+    /**
+     * Render function of component.
+     *
+     * @returns {ReactElement}
+     */
+    render() {
+        return (
+            /* eslint-disable */
+            <svg width='24' height='24' viewBox='0 0 24 24' focusable='false' role='presentation'>
```
saghul: can't we use some loader for this? instead of embedding the SVG in render(), that is.

AkshitKrNagpal: I think it might be possible with [svg-inline-loader](https://github.com/webpack-contrib/svg-inline-loader). I'll give it a try. :smile:

AkshitKrNagpal:
> Do you think we can achieve this?

Looks like we can this way.

saghul:
Nice work! I'm going to play around with the structure a bit and see what feels best. I'm not going to bother you with making these changes, because I may change my mind several times :-P

saghul:
I made a slight simplification since I think we don't need such a generic feature at this time, but really nice work @AkshitKrNagpal! :+1:

PR #57, opened 2018-06-21
Improve dragging experience

saghul:
The entire page is draggable, except:

- buttons
- any input type
- the filmstrip

### DIFF
--- a/app/features/conference/components/Conference.js
+++ b/app/features/conference/components/Conference.js
@@ -168,7 +168,6 @@ class Conference extends Component<Props, *> {
         setupWiFiStats(iframe);
 
         this._api.on('readyToClose', () => this._navigateToHome());
-
         this._api.on('videoConferenceJoined',
             (conferenceInfo: Object) =>
                 this._onVideoConferenceJoined(conferenceInfo));
@@ -208,6 +207,7 @@ class Conference extends Component<Props, *> {
      * @returns {void}
      */
     _onVideoConferenceJoined(conferenceInfo: Object) {
+        setupDragAreas(this._api.getIFrame());
 
         this._setAvatarURL(this.props._avatarURL);
         this._setEmail(this.props._email);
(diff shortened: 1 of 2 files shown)


app/index.html
```diff
@@ -8,6 +8,10 @@
         body {
             -webkit-app-region: drag
         }
+        button,
+        input[type] {
```
AkshitKrNagpal: Why `input[type]` ?
Why not just `input`?

saghul: It matches only inputs which have a type attribute. TBH I'm not sure that's enough, but it seems to work well in practice 😄

app/features/conference/components/Conference.js
```diff
@@ -208,6 +207,17 @@ class Conference extends Component<Props, *> {
      * @returns {void}
      */
     _onVideoConferenceJoined(conferenceInfo: Object) {
+        // Inject some style into the iframe so everything except the filmstrip,
+        // buttons or any input is draggable.
+        const iframe = this._api.getIFrame();
+        const injectStyle = document.createElement('style');
+
+        injectStyle.type = 'text/css';
```
hristoterezov: Isn't it going to be better if you put this code into separate function?

I think this is good enough for temporary solution but isn't it going to be better if we implement more generic way to inject styles into meet from the iframe api?

app/index.html
hristoterezov: But something is not working if we use just input?

saghul: I don’t know, we can change it.

app/features/conference/components/Conference.js
saghul: I can move to a function if that feels better.

As for having an API, I don’t disagree, but that isn’t orthogonal to this. We can use that, once / if we have it.

saghul:
What I’m trying to solve is providing a good dragging expresidente in the entire window. She the entire window is draggable, you cannot select text in a text field, for example. Also, you can’t scroll the chat or the filmstrip by clicking on the scrollbar. That is what this PR fixes.

app/features/conference/components/Conference.js
hristoterezov: sure!

saghul:
@hristoterezov Updated, PTAL!

PR #60, opened 2018-06-23
Added defaultServerURL to settings

### LINKED ISSUE #46: settings: add ability to select custom server

AkshitKrNagpal:
Resolves #46 .


app/features/settings/components/SettingsDrawer.js
```diff
@@ -95,7 +102,15 @@ class SettingsDrawer extends Component<Props, *> {
...
                                 value = { this.props._email } />
                         </form>
-                    </ProfileContainer>
+                        <form onSubmit = { this._onDomainFormSubmit }>
+                            <FieldText
+                                label = 'Default Domain'
+                                onBlur = { this._onDomainBlur }
+                                shouldFitContainer = { true }
+                                type = 'text'
+                                value = { this.props._defaultDomain } />
```
saghul: please add a placeholder which uses the value in `config.defaultDomain`.

app/features/conference/components/Conference.js
```diff
@@ -81,7 +85,8 @@ class Conference extends Component<Props, *> {
     componentDidMount() {
         const parentNode = this._ref.current;
         const room = this.props.match.params.room;
-        const domain = this.props.match.params.domain || config.defaultDomain;
+        const domain = this.props.match.params.domain
+            || this.props._defaultDomain;
```
saghul: This needs to also consider `config.defaultDomain` in case the user entered nothing.

app/features/settings/actionTypes.js
```diff
@@ -8,6 +8,16 @@
...
 
+/**
+ * The type of (redux) action that sets the default domain.
+ *
+ * {
+ *     type: SET_DEFAULT_DOMAIN,
+ *     defaultDomain: string
+ * }
+ */
+export const SET_DEFAULT_DOMAIN = Symbol('SET_DEFAULT_DOMAIN');
```
saghul: let's call this entire thing a "server URL". this is consistent with jitsi meet on mobile.

app/features/settings/components/SettingsDrawer.js
```diff
@@ -73,7 +80,7 @@ class SettingsDrawer extends Component<Props, *> {
                 onBackButton = { this._onBackButton }
                 primaryIcon = { <Logo /> } >
                 <DrawerContainer>
-                    <ProfileContainer>
+                    <SettingsContainer>
```
saghul: Thanks for changing this!

app/features/conference/components/Conference.js
AkshitKrNagpal: The reducer actually sets that as default value for defaultDomain. 
You must be experiencing this because of redux-persist. 

you'll need to flush previous values.

They way I do it is.
Inspect element -> React Tab -> Expand Root and Provider -> Select PeristorGate -> in right side you will see props -> Expand persistor -> You will see functions -> right click on flush function. and select "execute function"

saghul: That will only work the first time. Go ahead and edit the field in settings and set it empty.

app/features/settings/actionTypes.js
AkshitKrNagpal: Okey. I'll update this.

app/features/conference/components/Conference.js
AkshitKrNagpal: @saghul 
I guess this can be fixed with valid server checking since "blank" will not be a valid server. Since we currently don't have error display in Conference component, this could be handled by that also.
But keeping the current functionality in mind, I think this is good to go for this PR. WDYT?

app/features/settings/actionTypes.js
AkshitKrNagpal: @saghul 
On second thought, aren't we only storing domain? 
Why should we call this server URL?
Since we can't support HTTP here we don't need full URL just domain would be sufficient.

saghul: We support deploying in a subdirectory, so something like https://example.com/meet/ should be valid too.

We can strip the scheme if specified, and force it to https, that’s ok.

app/features/conference/components/Conference.js
saghul: If the user enters a URL and deletes it later it must continue to work. I’ll test again later today 👍

AkshitKrNagpal: > If the user enters a URL and deletes it later it must continue to work.

I guess then, I should add something like this in middleware. 
If its blank, then `dispatch(setDefaultServerURL(config.defaultServerURL)`.
WDYT?

app/features/settings/actionTypes.js
AkshitKrNagpal: Oh. In that case, I'll update this.

app/features/conference/components/Conference.js
AkshitKrNagpal: Okay. Then I'll set it `undefined`.

saghul:
@AkshitKrNagpal So, I tested it again, and I think we need to change how we handle this internally. Up until now we have had the concept of "domain", but in reality it's a URL. I'm sorry for not pointing out this earlier.

I think not requiring the URL scheme sounded like a good idea but may be detrimental now. We should support `https://my.server.com` as the server URL. If we can support not specifying the scheme, without too much trouble, then great, but otherwise we need to make it mandatory.

In order to achieve this, some naming adjustments are also necessary:

- `config.defaultDomain` -> `config.defaultServerURL`
- `SET_DEFAULT_SERVER_URL` -> `SET_SERVER_URL`

The concept of the "default server URL" is what's in `config`, what the user specifies is just "server URL".

PS: I know this part is taking quite some time, but it's crucial that we get this right, and you are doing an awesome job so far!

AkshitKrNagpal:
> The concept of the "default server URL" is what's in config, what the user specifies is just "server URL".

No it isn't. It is default server URL , ie, the url of server to go if full url is not given. And just room name is given. And config has the initial value of default server url.

>  PS: I know this part is taking quite some time, but it's crucial that we get this right, and you are doing an awesome job so far!

Don't worry. 😄

saghul:
> No it isn't. It is default server URL , ie, the url of server to go if full url is not given. And just room name is given. And config has the initial value of default server url.

I see why you'd think that way, and it's not incorrect either. I just don't want to see "default" everywhere. Think of it as setting **your** server, and then just using a different one if you specify the full URL.

We are overloading the term "default" a bit, indeed, but hopefully I could explain well enough :-)

AkshitKrNagpal:
>  I just don't want to see "default" everywhere.

That does make sense. Okay. I will update this. 😄 

> config.defaultDomain -> config.defaultServerURL

This too. 😄

AkshitKrNagpal:
For domain -> server

How about this?

`config.defaultServerURL='https://meet.jit.si'`
Whatever we have in the input field, we'll remove everything before `://` (nothing if this not found) and append `https://`. Now what we have will be serverURL. 

@saghul

saghul:
> For domain -> server
> 
> How about this?
> 
> config.defaultServerURL='https://meet.jit.si'
> Whatever we have in the input field, we'll remove everything before :// (do nothing if this not found) and append https://. Now what we have will be serverURL.

Sounds exactly what's in my mind! :-)

AkshitKrNagpal:
> Sounds exactly what's in my mind! :-)

Cool. I'll implement this. 👍 
ASAP.

saghul:
@AkshitKrNagpal ping. domain -> server URL change landed.

AkshitKrNagpal:
Ping @saghul. PTAL.

saghul:
Absolutely wonderful job here @AkshitKrNagpal :tada: :tada: :tada: This is a really important milestone in this project, and we couldn't be happier with the result so far.

Please take a moment to reflect on this, your work is making a huge impact in this project, and we are very lucky to have such a talented developer working with us thanks to GSoC. :beers:

AkshitKrNagpal:
@saghul 
Thank you so much. 😄

PR #61, opened 2018-06-23
Replaced getAvatarURL from js-utils

### LINKED ISSUE #54: avatar: share code with jitsi-meet

AkshitKrNagpal:
Resolves #54.
Depends on https://github.com/jitsi/js-utils/pull/5


webpack.config.js
```diff
@@ -12,7 +12,7 @@ const commonConfig = {
     module: {
         rules: [
             {
-                exclude: /(node_modules)/,
+                exclude: /node_modules(?!\/js-utils)/,
```
AkshitKrNagpal: This tells Webpack to not exclude js-utils and use babel-loader.

package.json
```diff
@@ -44,7 +44,7 @@
     "electron-window-state": "4.1.1",
     "history": "4.7.2",
     "jitsi-meet-electron-utils": "github:jitsi/jitsi-meet-electron-utils#1972c3bf0884ace68eb496894dabae593d6dbf49",
-    "js-md5": "0.7.3",
+    "js-utils": "github:akshitkrnagpal/js-utils#avatar",
```
AkshitKrNagpal: I'll update this when PR in js-utils is merged.

saghul: :+1:

webpack.config.js
saghul: Why do we need the babel loader for it?

AkshitKrNagpal: `js-utils` needs to use babel-loader. 
I got errors so I checked how `lib-jitsi-meet` uses it. I found something similar. https://github.com/jitsi/lib-jitsi-meet/blob/7b39059fba68240ac930af30cc99b655a3ba7edf/webpack.config.js#L46

package.json
hristoterezov: @AkshitKrNagpal the js-utils PRs is merged!

AkshitKrNagpal:
Ping @hristoterezov.

PR #66, opened 2018-06-24
Add keeyboard shortcut to open settings drawer

AkshitKrNagpal:
@saghul 
Does 'command+s' and 'ctrl+s' sound good?


AkshitKrNagpal:
For mac, `command + ,` looks good.

For others, I don't think there is a general trend for settings shortcut.

saghul:
LGTM, but I found a little nuisance: it doesn't work if the room textfield is focused, which is the default when the app is loaded. Is there anything we can do here?

AkshitKrNagpal:
@saghul 
This does it. But importing it is a little mess.

PR #68, opened 2018-06-24
Add "about" link to help menu

saghul:


app/features/navbar/components/HelpButton.js
```diff
@@ -101,10 +77,10 @@ export default class HelpButton extends Component< *, State> {
                 position = 'right bottom'
                 trigger = { <HelpIcon /> }>
                 <Group heading = 'Help'>
-                    <Item onActivate = { this._onTermsOptionClick }>
+                    <Item onActivate = { this._onTerms }>
```
AkshitKrNagpal: I guess it should be at least `onTermsClick`?

saghul: I went with the shorter version, WDYT?

AkshitKrNagpal: It just doesn't feel like a listener function anymore. 😅

saghul: Yeah, you’re right. I’ll fix that!

saghul:
@AkshitKrNagpal updated, PTAL.

app/features/navbar/components/HelpButton.js
```diff
@@ -97,9 +76,20 @@ export default class HelpButton extends Component< *, State> {
                 onOpenChange = { this._onIconClick }
                 position = 'right bottom'
                 trigger = { <HelpIcon /> }>
-                <Item onActivate = { this._onTermsOptionClick }>Terms</Item>
-                <Item onActivate = { this._onPrivacyOptionClick }>Privacy</Item>
-                <Item onActivate = { this._onSendFeedback }>Send Feedback</Item>
+                <Group heading = 'Help'>
```
AkshitKrNagpal: Sorry I didn't at the UI previously.
Should we really add this? The UI looks a little bit weird to me.

AkshitKrNagpal:
Thanks @saghul 
LGTM minus group thing. :smile:

app/features/navbar/components/HelpButton.js
saghul: It's consistent with other UI I've seen using AtlasKit.

AkshitKrNagpal: @saghul 
Hmm. Yeah, it is. I just looked at stride and bitbucket.
Maybe it looks weird to me because it is centre-aligned. 
Totally your call. :smile:

PR #69, opened 2018-06-25
Changed domain to serverURL

AkshitKrNagpal:
Ping @saghul.
PTAL. 

Since there is no way of passing server URL as match param I ended up upgrading to memory router and pass room and server URL as the state.
Previous logic was made for a domain without keeping in mind that the server can have a path, this fixes that also.


saghul:
Landed with a small change: avoid navigation if room is empty. Nice work @AkshitKrNagpal! This feels much better aligned with the PR for handling different server URLs now.

Issue #47, opened 2018-06-18
build: add app icon


saghul:

saghul:
https://www.electron.build/icons

akshitkrnagpal:
@saghul 
Can you close this too, please? 😄

saghul:
Roger that!

PR #74, opened 2018-07-02
WIP Added electron builder to generate binaries

### LINKED ISSUE #48: build: generate binaries
- [ ] macOS: a DMG
- [ ] Windows: an installer (msi / exe)
- [ ] Linux: an AppImage

akshitkrnagpal:
Will resolve #48, #47.

It can successfully build these on the same platforms
- [x] Mac (DMG)
- [x] Linux (AppImage)
- [x] Windows (EXE)

To Do
- [ ] Add documentation

Optional
- [ ] Create Graphics (Sidebar Image for Win and DMG background for Mac)


akshitkrnagpal:
> Can we customize the binary name? 

Yeah, there a whole lot of customizations. I am currently testing it on Linux and will do that.

akshitkrnagpal:
>  It's right now taken from the package.json name so it's "jitsi-meet-electron", but we want it to be "jitsi-meet" on Linux and Windows, and "Jitsi Meet" on macOS.

I have updated this.

> Amazing, I just tested it on Linux and the AppImage works just fine!

Did it previously show Jitsi icon? Which distro were you using?

saghul:
> Did it previously show Jitsi icon? Which distro were you using?

I don't remember the icon part, I'll test it again when I get home. I was using Fedora.

saghul:
@akshitkrnagpal Can you please rebase this, now that the toggles landed? Thanks!

akshitkrnagpal:
Rebased. :smile:

saghul:
@akshitkrnagpal I tested again (Fedora). The AppImage is generated and works correctly, but the icon is not showing correctly.

akshitkrnagpal:
> The AppImage is generated and works correctly,

Does it add a desktop entry?

saghul:
I didn’t enable desktop integration. The broken icon is on the task switcher (Alt-Tab) sorry for not being specific.

package.json
```diff
@@ -2,16 +2,65 @@
...
+      "category": "public.app-category.video",
+      "artifactName": "${productName}-${version}-${arch}.${ext}"
+    },
+    "linux": {
+      "artifactName": "${productName}-${version}-${arch}.${ext}",
+      "category": "VideoConference;AudioVideo;Audio;Video;Network",
+      "description": "Jitsi Meet Desktop App",
+      "executableName": "jitsi-meet",
+      "target": [
+        "AppImage",
```
saghul: let's do only AppImage for now, since that works everywhere.

package.json
```diff
@@ -26,7 +75,7 @@
     "electron",
     "jitsi-meet"
   ],
-  "author": "",
+  "author": "Akshit Kr Nagpal <nagpalkrakshit@gmail.com>",
```
saghul: please set this to "Jitsi Team <support@jitsi.org>"

package.json
```diff
@@ -2,16 +2,65 @@
...
+  "build": {
+    "appId": "org.jitsi.jitsi-meet",
+    "productName": "Jitsi Meet",
+    "files": [
+      "**/*",
+      "resources",
+      "!app",
+      "!main.js"
+    ],
+    "copyright": "",
```
akshitkrnagpal: @saghul 
Can you please suggest the value for this also. 😄

package.json
akshitkrnagpal: Yeah. Sure

package.json
akshitkrnagpal: I was thinking let's do as many as possible with minimal configuration.
Since they are already working, I don't see a reason not to include them. WDYT?

akshitkrnagpal:
> I'll test on Windows ASAP.

Windows will give an error in webpack build. It depends on fd8ae7b
Please merge other PR first.

package.json
saghul: If we publish those binaries we set the expectation that they work. However, realistically, we are not going to test them all,plus we open the door for questions like "why don't you provide RPM packages?". So, I'd say, let's start small, just AppImage, which works everywhere, and add more in the future if necessary. Adding stuff is easy, removing stuff once it's out there... not so much.

akshitkrnagpal: >  we open the door for questions like "why don't you provide RPM packages?"

Yeah, That will be annoying. :sweat_smile: 

>  just AppImage, which works everywhere

If that's the reason, how about a snap package? 
I find snap packages more easy to install. (That may be because I am using Ubuntu)
Also, if I don't specify target it builds AppImage and snap package. Maybe I can remove this and let `electron-builder` use default?

saghul: > If that's the reason, how about a snap package?
> I find snap packages more easy to install. (That may be because I am using Ubuntu)

Then the Fedora people will say: "why not Flatpak?" It's the same thing all over again :-)

> Also, if I don't specify the target, it builds AppImage and snap package. Maybe I can remove this and let electron-builder use default?

I'd rather be explicit about what we want: AppImage and nothing else.

> Other problems with AppImage
> 
>     Have to move .AppImage to organise them
>     If we move it after desktop integration. it will break.

Move it, how? You mean move the file location? If that's the case, that is not a problem we create, but a problem inherent to AppImage, so I'd say we put a line in the documentation and call it a day :-)

akshitkrnagpal: Cool. I have updated this.

akshitkrnagpal:
I have rebased and updated this.
If you haven't you can check on windows now. 😄

saghul:
I won't be able to get access to a Windows machine until Friday, but I'll check the rest in the meantime! :muscle:

PR #76, opened 2018-07-04
Minor Fixes

akshitkrnagpal:


main.js
```diff
@@ -116,6 +117,9 @@ function createJitsiMeetWindow() {
     jitsiMeetWindow.on('closed', () => {
         jitsiMeetWindow = null;
     });
+    jitsiMeetWindow.once('ready-to-show', () => {
+        jitsiMeetWindow.show();
```
akshitkrnagpal: This is a common practice in Electron apps.
There is a minor time difference in rendering of the app. During that time a white-ish blank screen is shown. 
This time is larger in slow systems. This will fix that.

saghul: Excellent, thank you Akshit!

PR #81, opened 2018-07-08
WIP Added error flag when unable to join conference

akshitkrnagpal:


akshitkrnagpal:
@saghul 

> I think it might be very useful to generalise this and have a feature which keeps and manages a list of flags. That way we can show a new flag when a new feature is introduced, maybe plug Stride or something like that.

That sounds great. And would be great for announcing new features. (as you said)

But actually, I was having doubts whether we should use Flag for this error msg. Maybe just form validation would look better in this case?
We'll show an error with the input field. WDYT?

saghul:
> Maybe just form validation would look better in this case?
> We'll show an error with the input field. WDYT?

Let's give it a try and see how it feels like. Good idea Akshit!

akshitkrnagpal:
I guess I'll have to create custom style for message. 
screen shot 2018-07-10 at 1 20 48 am

This looks good. 
screen shot 2018-07-10 at 1 20 35 am

saghul:
The second one looks good indeed. Now, can the position of the first message be set? If not, let's go with the second options.

akshitkrnagpal:
> Now, can the position of the first message be set?

Atlaskit doesn't have that option (or it's not documented)

> If not, let's go with the second options.

Currently, it looks like the first image when the input box is focused and like second if it not.
Maybe just removing the message would be enough? WDYT?

saghul:
> Currently, it looks like the first image when the input box is focused and like second if it not.
> Maybe just removing the message would be enough? WDYT?

Sounds good!

akshitkrnagpal:
@saghul 
Updated.

Issue #67, opened 2018-06-24
Handle URL parameters


saghul:
A user may enter the following URL: `https://meet.jit.si/fooo#config.p2p.enabled=false`

We need to convert all the hash parameters into config overrides which the iframe API takes (right @hristoterezov ?).

AkshitKrNagpal:
I guess then we should have a function in `utils` that parses everything and returns an object with all the options.

Also, I guess we can have these settings for default server as well or do you think they should do this in their server config?

saghul:
We currently have (or rather, will have) 2 settings: startWithAudioMuted and startWithVideoMuted. Those should take precedence over the stuff in the URL. So yeah, I think we should parse the URL and extract the hash parameters, and apply them together with those 2 when we join.

Let's discuss this is tomorrow's call! This goes after the start muted toggles anyway, so no problem.

agraef:
I'd really like to give this a bump and a big thumbs up. For me, it is especially important to override audio-related options to get better audio quality when transmitting music (I'm teaching computer music courses via Jitsi Meet to my students at university). I currently have to resort to using Jitsi Meet in Chromium to make that work.

Can we please get this? Like really soon now? ;-)

I'd be willing to have a look at the sources myself if someone points me to the right place.

chamerling:
Hello,

This has been addressed in https://github.com/jitsi/jitsi-meet-electron/pull/286 and available since v2.1.1

saghul:
Indeed!

agraef:
@chamerling Thanks for the heads-up! I actually saw your PR and already thought that this might fix it, so I got quite excited when the upgrade to version 2.3.0 landed on my system this morning. :)

However, giving it a quick whirl adding the parameters that I usually employ, `#config.p2p.enabled=false&config.disableAP=true&config.disableAEC=true&config.disableNS=true&config.disableAGC=true&config.disableHPF=true&config.stereo=true`, I still noticed some artifacts on the receiving end, so I wasn't sure whether that's actually supposed to work now. But I'll give it another try. Sometimes it's hard to hear (or rather not just imagine) the artifacts when you're looking for them...

agraef:
[...]

PR #82, opened 2018-07-10
Handle Tenant name in Server URL

akshitkrnagpal:


akshitkrnagpal:
Ping @saghul 

I wonder if deleting `global._babelPolyfill` like that would have any unforeseen bugs. Any thought?

webpack.config.js
```diff
@@ -18,7 +18,12 @@ const commonConfig = {
                     presets: [
                         [
                             require.resolve('babel-preset-env'),
-                            { modules: false }
+                            {
+                                modules: false,
+                                targets: {
+                                    electron: '2.0.3'
```
saghul: can we say something like >= 2 here?

akshitkrnagpal: @saghul 
Nops. It throws an error. I already tried '>=2.0.0' and '2.0.x' too. 
It says 'Versions must be in semver format (major.minor.patch).
IDK why doesn't it take those as valid.

I tried using `process.versions.electron` too but the electron is not initialised at that point.

Should I require `package.json` as JSON object if you want it to be automatic?

saghul: > Should I require package.json as JSON object if you want it to be automatic?

Yeah, that sounds good!

akshitkrnagpal: Done.

app/features/utils/functions.js
```diff
@@ -5,6 +5,29 @@
...
 import config from '../config';
 
+/**
+ * Finds the external_api.js for the given Server URL. It also handles if the
+ * server URL contains tenant name.
+ *
+ * @param {string} serverURL - Jitsi Meet Server URL.
+ * @returns {Promise} - Promise which returns the external_api.js url.
+ */
+export async function findExternalApiURL(serverURL: string) {
```
saghul: Something is wrong here. If I use some incorrect URL it never returns an error. Try https://saghul.net/foo/bar

Also, for some reason I get a weird result with https://example.com/foo/bar

saghul:
I get to an example.com page and to an infinite loop. I’ll test again.

Yeah, if we are supporting tenants I’d say let’s do it all the way.

akshitkrnagpal:
>  I’ll test again.

Okay

> Yeah, if we are supporting tenants I’d say let’s do it all the way.

Cool. I'll update this.

akshitkrnagpal:
> I get to an example.com page and to an infinite loop. I’ll test again.

Oooh. You must have put `http` instead of `https`. I'll handle that and update this.

saghul:
I actually skipped the scheme completely.

akshitkrnagpal:
> I actually skipped the scheme completely.

Yeah, I found I was missing a catch. Sorry. 😅 
Fixed and updated in SettingsDrawer.
@saghul

app/features/utils/functions.js
```diff
@@ -3,7 +3,33 @@
...
+    const apiURL = getExternalApiURL(serverURL);
+
+    const response = await fetch(apiURL, { method: 'HEAD' });
+
+    if (response.ok) {
+        return apiURL;
+    }
+    // eslint-disable-next-line no-param-reassign
+    serverURL = serverURL.replace(/https?:\/\//, '');
```
saghul: why do we need to do this? we are adding the scheme in normalizeServerURL (called from getExternalApiURL) and then removing it here?

saghul:
Works OK, so I'm ready to merge once we clarify that little thing. Ideally I'd like to avoid parsing the URL multiple times.

app/features/utils/functions.js
akshitkrnagpal: We need to strip it here because we need to calculate  `serverURL.lastIndexof('/')`.
If we don't it will find `/` in `https://`.

saghul: can we use a URL object in fetch() ? If so, we could do that in url.pathname.

akshitkrnagpal: > can we use a URL object in fetch() ? If so, we could do that in url.pathname.

I see what you are saying. It might look cleaner. I'll try that.

akshitkrnagpal: > can we use a URL object in fetch() ? If so, we could do that in url.pathname.

@saghul 
Using the URL object is getting confusing and it does not reduce any amount of code. 
It is also forcing to create a new object and I actually want to avoid that.

saghul: Hum, then let's try something else: process the URL and create all variations of it beforehand in an array, then process the items in turns, one at a time (from longest to shortest). This way we'd only fiddle with the URL parsing once, I think.

akshitkrnagpal: You mean don't make it recursive?

IMHO, (and I am not being lazy 😄 ) it will just increase amount of code and complexity.

saghul: I'd like us to not do and undo the HTTPS thing on the URL for each turn, please. If you have another option in mind, I'll be happy to hear it!

akshitkrnagpal: @saghul 
We aren't undoing it. Is there a confusion?
The `getExternalApiURL` value goes to the `apiURL`.
And we are removing `https://` from `serverURL` and we only really need it to be in the first call. Not all recursive calls. But it checks whether to do it or not automatically.

akshitkrnagpal:
@saghul 
PTAL. I avoided that.
It works just like that.

app/features/utils/functions.js
```diff
@@ -3,7 +3,34 @@
...
+    serverURL = normalizeServerURL(serverURL);
+    const apiURL = getExternalApiURL(serverURL);
+
+    const response = await fetch(apiURL, { method: 'HEAD' });
+
+    if (response.ok) {
+        return apiURL;
+    }
+
+    const start = 'https://'.length;
```
saghul: I really don't like this. We are not leveraging the work URL objects already do for us. They are cheap, let's use them. Also I just tested it and browsers seem to update href when pathname is changed. Maybe we need to use the whatwg-url polyfill, that's ok.

I know this is tedious, and I'm sorry, but I want to get it right, and I think we are not there yet. If you want to leave this and come back to it later, that's ok too!

akshitkrnagpal: > I know this is tedious, and I'm sorry, but I want to get it right,

Don't say sorry. I want to get it right too. :smile: 

> We are not leveraging the work URL objects already do for us. They are cheap, let's use them. Also I just tested it and browsers seem to update href when pathname is changed. Maybe we need to use the whatwg-url polyfill, that's ok.

Maybe I was doing something wrong. I'll try again. 


@saghul 
Let's do this. We'll revisit this after we fix #67 `Handle URL Parameters.` issue.
Maybe I will have more clarity about how to get it right as we'll need to parse complete url there.
:smile:

saghul: Sounds good! Thanks man!

akshitkrnagpal:
We decided not to support Tenants. 
Hence closed this.

PR #85, opened 2018-07-13
Misc

akshitkrnagpal:
- [x] Added electron target to babel-preset-env c05f9e8
- [x] Fix: Made host parsing compatible for servers deployed in subdirectory 6eb5a70
- [x] Force single instance application ef0380d


saghul:
Thanks or the fixes @akshitkrnagpal! Can you please elaborate on what the single instance thing does? Does that function return a boolean?

akshitkrnagpal:
> Can you please elaborate on what the single instance thing does? 

Glad to. 
If you try to open Jitsi Meet application when there is already a Window present it executes the callback we provide to `makeSingleInstance` function. 

> Does that function return a boolean?

Yes, it does. 

[Link for docs](https://electronjs.org/docs/all#appmakesingleinstancecallback)

PS - You can test it by executing `npx electron ./build/main.js` in a new terminal window. (Assuming you already have `npm start` in one window)

PR #100, opened 2018-07-27
Recent list improvements

saghul:
screen shot 2018-07-27 at 10 20 13

### DIFF
--- a/app/features/recent-list/components/RecentList.js
+++ b/app/features/recent-list/components/RecentList.js
@@ -6,7 +6,12 @@ import { connect } from 'react-redux';
 import type { Dispatch } from 'redux';
 import { push } from 'react-router-redux';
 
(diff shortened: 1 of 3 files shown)


app/features/recent-list/components/RecentList.js
```diff
@@ -64,14 +69,17 @@ class RecentList extends Component<Props, *> {
             <ConferenceCard
                 key = { conference.startTime }
                 onClick = { this._onNavigateToConference(conference) }>
-                <TruncatedText>
+                <ConferenceTitle>
```
akshitkrnagpal: I wonder if the right thing to do should pass props? See [link](https://www.styled-components.com/docs/basics#passed-props)

`<TruncatedText bold = { true }> { conference.room } </TruncatedText>`

But I guess extending would be simpler?

akshitkrnagpal:
Looks Great. !!

PR #102, opened 2018-08-01
WIP Added onboarding

### LINKED ISSUE #64: Add onboarding on the first run
http://atlaskit.atlassian.com/packages/core/onboarding

- tell the user what to enter in the room field
- expand and explain settings

akshitkrnagpal:
Resolves #64 

TODO
- [x] Fix onboarding. 🤕 
- [x] Welcome Modal with actions (Start Tour and Skip)
- [x] Add onboarding modal image
- [x] Persist


app/features/onboarding/constants.js
```diff
@@ -0,0 +1,25 @@
...
+
+export const onboardingSteps = [
+    'conference-url',
+    'settings-drawer-button',
+    'name-setting',
+    'email-setting',
+    'server-setting'
+];
+
+export const onboardingComponents = {
```
akshitkrnagpal: As we have to persist which onboarding components have been shown already, we can't directly use components. Hence we have to map them with strings.

app/features/app/components/App.js
```diff
@@ -50,3 +51,5 @@ export default class App extends Component<*> {
         );
     }
 }
+
+export default connect()(App);
```
saghul: Is this needed now?

app/features/onboarding/actionTypes.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * The type of (redux) action that hides active onboarding.
+ *
+ * {
+ *     type: HIDE_ACTIVE_ONBOARDING,
+ *     parentComponent: string
```
saghul: How about calling this "section" instead?

app/features/welcome/components/Welcome.js
```diff
@@ -104,12 +128,31 @@ class Welcome extends Component<Props, State> {
...
             </Page>
         );
     }
 
+    /**
+     * Render active onboarding component.
+     *
+     * @returns {void}
+     */
+    _renderActiveOnboardingComponent() {
```
saghul: I think this could be put into a `OnboardingContainer` or just `Onboarding` component, which gets a prop with the section name. That way it can be reused in both locations.

app/features/settings/components/SettingsDrawer.js
```diff
@@ -63,6 +74,24 @@ class SettingsDrawer extends Component<Props, *> {
...
+     *
+     * NOTE: It autonatically checks if the onboarding is shown or not.
+     *
+     * @returns {void}
+     */
+    componentDidUpdate() {
+        if (this.props.isOpen) {
+
+            // TODO - Find a better way for this.
+            // Delay for 1 second to let drawer open.
```
saghul: Is one full second necessary? Or just a 100ms setTimeout would do?

app/features/settings/components/SettingsDrawer.js
```diff
@@ -63,6 +74,24 @@ class SettingsDrawer extends Component<Props, *> {
...
 
+    /**
+     * Start Onboarding once component is mounted.
+     *
+     * NOTE: It autonatically checks if the onboarding is shown or not.
+     *
+     * @returns {void}
+     */
+    componentDidUpdate() {
+        if (this.props.isOpen) {
```
saghul: can you check if `!prevProps.isOpen` here? (just playing safe)

app/features/settings/components/SettingsDrawer.js
akshitkrnagpal: Drawer animation is 220ms. So I guess we at least need that much.
On fast devices, 250ms works just fine. But on a little slow device 300ms sometimes does it correctly and sometimes the first onboarding created the problem which I mentioned earlier.

WDYT, how much should we set?

app/features/onboarding/actionTypes.js
akshitkrnagpal: Done

app/features/welcome/components/Welcome.js
akshitkrnagpal: Nice catch.
I tried that when we didn't have sections for onboarding but I guess I just forgot. :smile:

app/features/settings/components/SettingsDrawer.js
akshitkrnagpal: Done

app/features/settings/components/SettingsDrawer.js
saghul: Let's do 500, and please put that 220 in the comment so we remember :-)

app/features/app/components/App.js
```diff
@@ -50,3 +50,5 @@ export default class App extends Component<*> {
         );
     }
 }
+
+export default App;
```
saghul: Can you please undo this change altogether? It's just not needed so it only adds "nose" to the PR.

saghul:
Looking good @akshitkrnagpal ! Just those 2 items missing and we got this! (I'll make copy edits after you're done to avoid force-pushing)

akshitkrnagpal:
@saghul 
I think this is done except Modal Image.

akshitkrnagpal:
- [x] Updated Dependencies

Issue #103, opened 2018-08-08
Add ability to watch / autoreload in development mode


saghul:
After running `npm start` it would be nice to be able to run `npm watch` on another terminal and reload the app when files change.

Candidate helper package: https://github.com/yan-foto/electron-reload

PR #108, opened 2018-08-16
Add ability to watch / autoreload in development mode

saghul:
Fixes: https://github.com/jitsi/jitsi-meet-electron/issues/103

### DIFF
--- a/main.js
+++ b/main.js
@@ -21,9 +21,14 @@ autoUpdater.logger = require('electron-log');
 autoUpdater.logger.transports.file.level = 'info';
 
 /**
- * Load debug utilities (don't open the DevTools window by default though).
+ * When in development mode:
+ * - Load debug utilities (don't open the DevTools window by default though)
+ * - Enable automatic reloads
  */
-require('electron-debug')({ showDevTools: false });
+if (isDev) {
+    require('electron-debug')({ showDevTools: false });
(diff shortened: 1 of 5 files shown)


main.js
```diff
@@ -25,6 +25,14 @@ autoUpdater.logger.transports.file.level = 'info';
  */
 require('electron-debug')({ showDevTools: false });
 
+/**
+ * Enable automatic reloads in development mode.
+ */
+
```
akshitkrnagpal: style: spurious empty line :stuck_out_tongue_closed_eyes:

main.js
```diff
@@ -25,6 +25,14 @@ autoUpdater.logger.transports.file.level = 'info';
...
+if (isDev) {
+    require('electron-reload')(path.join(__dirname, 'build'));
```
akshitkrnagpal: How about moving `require('electron-debug')` to this block too?

akshitkrnagpal:
Hey @saghul 
It doesn't automatically open window?

main.js
saghul: I taught you well :-)

akshitkrnagpal:
>  You mean there is a way to do this in a single command?

Yeah. I got it working. But that would only work for react stuff.

akshitkrnagpal:
LGTM.. !! 😄

PR #110, opened 2018-08-20
Add placeholder for conference field

saghul:
Also, create a couple of auxiliary functions so we save some indentation
in the main render function.


saghul:
@akshitkrnagpal You're not wrong, but I noticed it too late, so I went for a single commit then 🤷‍♂️

PR #137, opened 2018-10-10
Fix join regression

saghul:


saghul:
@hristoterezov I f*cked up last release. PTAL and let's make a quick release.

PR #166, opened 2019-05-06
chore(deps): bump jitsi-meet-electron-utils to v0.6.3

virtuacoplenny:


package.json
```diff
@@ -125,7 +125,7 @@
     "concurrently": "4.1.0",
     "css-loader": "2.0.1",
     "devtron": "1.4.0",
-    "electron": "3.0.13",
+    "electron": "4.1.5",
```
saghul: Why not 5, while we are at it?

virtuacoplenny: It's a little bit more work because robotjs doesn't work well with electron 5 without an update. So updating would require updating/rebasing the robotjs fork and updating utils' robotjs. I can do it but I almost would rather do the robotjs update in a separate commit.

saghul: agreed, you'll get extra bonus nachos if we can drop the fork and use upstream directly :-)

hristoterezov: @saghul We can't use the current upstream directly until https://github.com/octalmage/robotjs/pull/307 get merged. And since it is in review since 2017 I've lost hope...

saghul: Can you please follow up? You first sent the PR in june 2017, and your next message is one fukll year later. Sometimes a small push is needed ;-)

hristoterezov: I don't see a reason to follow up! Periodically there are people following up (the last one is from mid-march). Feel free to leave a comment there if you want!

saghul: You don't?! I really don't understand. If **you** don't care about your PR, who will? If we manage to push it upstream we'd get the advantage of the prebuilt binaries, for example, simplifying updates.

hristoterezov: To be clear:
 - I obviously understand the benefits of using the upstream. And clearly since I made the effort  to create the PR I also want to be able to use upstream.
 - You can check that I also have some comments/replies on the PR and I'm following its state. Yes there have been few months back in 2018 that I haven't been following it but this doesn't mean I don't care about this PR.

What I meant is:
 - You can see that on 10/3/2018 and on 03/15/2019 there are comments that are asking the owner of the project what is the status and he is not replying. I don't see how another comment like this from me a month later would make him merge it... 
 - Honestly I don't believe the owner of the project will ever take care of this PR since he doesn't have objections to the solutions and still 2 years after the PR was created there is no progress...

So what I'm saying is, if you disagree with me or if you have another strategy to get the PR merged feel free to apply it...

Issue #171, opened 2019-06-25
Option to disable always on top window


patrickpichler:
Hi,

is it possible to disable the always on top window? 
It is a bit annoying as it gets in my way most of the time.

saghul:
There currently is no option for it. Adding one is not out of tyhe question, but I don't think we have the time now, patches are welcome though!

saghul:
We just released a new version and the AoT window now has a X button so you can close it.

casphas:
Just tested version 2.0.0 of the electron app on Linux (Debian 10, XMonad window manager). The button to close the AoT window works but the window pops up again as soon as the main window loses focus the next time.
An option to turn off the AoT window entirely would be greatly appreciated. Or a pointer to where this might be added and I can attempt a patch.

aseeg:
I have the same problem when using the window manager i3wm: Switching to another workspace triggers the creation of the small window, which is stretched to cover the whole screen because of my standard settings and i3wm being a tiling window manager. I can then move the window onto another workspace/out of view, but as soon as I focus the main jitsi-meet-electron window, the small always-on-top-window vanishes, making it reappear as soon as I unfocus the main window (again stretched over my whole screen because of my i3wm standard settings).

I'd be happy about a way to disable the always-on-top window, or a way to stop it from disappearing when the main jitsi-meet-electron window gets the focus again.

Thank you for the hard work you put into Jitsi Meet and the Electron app.

cocoonkid:
I'd also wish for a way to disable it. It is useless 99% of the time for me.

A setting to chose between them on the client and server both would be great.

saghul:
Now that we have an advanced settings section, a PR adding a new setting there would be acceptable.

casphas:
I believe this has been implemented with PR #264.

Thanks!

Issue #168, opened 2019-05-29
feat: launch from URL with custom protocol like `jitsi://ROOM_ID`


tonytonyjan:
This would be a wonderful feature if it is possible.

I wonder if it is being planed in the future or it can not be done by Electron.

saghul:
Sure, why not! :-) We don't currently have much time to improve the Electron app, but if you send a PR I'll b e happy to review it.

ctramnitz:
why should the url be jitsi:// ? and why jitsi://test/<some_server_uri> (room & server reversed)?

The redirect to an installed app already uses org.jitsi.meet://server/room

saghul:
https://github.com/jitsi/jitsi-meet-electron/pull/263

PR #172, opened 2019-06-27
Implemented #168 and #171

gorance-teletrader:
This is updated version of app, this features are added.
#168 Add protocol to open app from browser
Example 1:                              jitsi://test
Example 2 (custom server):    jitsi://test/<some_server_uri>

#171 Disable window always on top
Create config
After changing the setting, reload of app is needed to applay changes
Note: due to lack of time, **auto reload** and **sync between stores** (browser and local file) are not implemented _yet_ :)

I tried to follow style of writing, but if something is not ok i will changed it.

One note, this is only tested on Windows 10
and could have bugs on different platforms

### DIFF
--- a/app/features/app/components/App.js
+++ b/app/features/app/components/App.js
@@ -26,6 +26,9 @@ class App extends Component<*> {
         super(props);
 
         document.title = config.appName;
+
+        this._listenOnProtocolMessages
+            = this._listenOnProtocolMessages.bind(this);
     }
 
     /**
@@ -54,6 +57,8 @@ class App extends Component<*> {
         );
     }
 
+    _listenOnProtocolMessages: (*) => void;
+
     /**
      * Handler when main proccess contact us.
      *
@@ -62,11 +67,10 @@ class App extends Component<*> {
      *
      * @returns {void}
      */
-    _listenOnProtocolMessages = (event, arg) => {
+    _listenOnProtocolMessages(event, arg) {
         // change route when we are notified
-        // eslint-disable-next-line no-invalid-this
         this.props.dispatch(push('/conference', arg));
-    };
+    }
 
     /**
      * Implements React's {@link Component#render()}.
--- a/app/features/app/components/App.js
+++ b/app/features/app/components/App.js
@@ -1,10 +1,12 @@
 // @flow
 
+import { ipcRenderer } from 'electron';
 import { AtlasKitThemeProvider } from '@atlaskit/theme';
 
 import React, { Component } from 'react';
 import { Route, Switch } from 'react-router';
-import { ConnectedRouter as Router } from 'react-router-redux';
+import { ConnectedRouter as Router, push } from 'react-router-redux';
+import { connect } from 'react-redux';
 
 import { Conference } from '../../conference';
 import config from '../../config';
@@ -14,17 +16,57 @@ import { Welcome } from '../../welcome';
 /**
  * Main component encapsulating the entire application.
  */
-export default class App extends Component<*> {
+class App extends Component<*> {
     /**
      * Initializes a new {@code App} instance.
      *
      * @inheritdoc
      */
-    constructor() {
-        super();
+    constructor(props) {
+        super(props);
 
         document.title = config.appName;
     }
+
+    /**
+     * Implements React's {@link Component#componentDidMount()}.
+     *
+     * @returns {void}
+     */
+    componentDidMount() {
+        // start listening on this events
+        ipcRenderer.on('protocol-data-msg', this._listenOnProtocolMessages);
+
+        // send notification to main process
+        ipcRenderer.send('renderer-ready');
+    }
+
+    /**
+     * Implements React's {@link Component#componentWillUnmount()}.
+     *
+     * @returns {void}
+     */
+    componentWillUnmount() {
+        // remove listening for this events
+        ipcRenderer.removeListener(
+            'protocol-data-msg',
+            this._listenOnProtocolMessages
+        );
+    }
+
+    /**
+     * Handler when main proccess contact us.
+     *
+     * @param {Object} event - Message event triggered by .
+     * @param {Object} arg - Object with two props same as Conference object.
+     *
+     * @returns {void}
+     */
+    _listenOnProtocolMessages = (event, arg) => {
+        // change route when we are notified
+        // eslint-disable-next-line no-invalid-this
+        this.props.dispatch(push('/conference', arg));
+    };
 
     /**
      * Implements React's {@link Component#render()}.
@@ -50,3 +92,5 @@ export default class App extends Component<*> {
         );
     }
 }
+
+export default connect()(App);
--- a/app/features/app/components/App.js
+++ b/app/features/app/components/App.js
@@ -12,6 +12,7 @@ import { Conference } from '../../conference';
 import config from '../../config';
 import { history } from '../../router';
 import { Welcome } from '../../welcome';
+import { createConferenceObjectFromURL } from '../../utils';
 
 /**
  * Main component encapsulating the entire application.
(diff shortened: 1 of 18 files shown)


main.js
```diff
@@ -211,3 +286,66 @@ app.on('window-all-closed', () => {
         app.quit();
     }
 });
+
+// remove so we can register each time as we run the app.
+app.removeAsDefaultProtocolClient(PROTOCOL_PREFIX);
+
+// If we are running a non-packaged version of the app && on windows
+if (process.env.NODE_ENV === 'development' && process.platform === 'win32') {
```
akshitkrnagpal: I think you can use `isDev` here.

saghul:
> Hi @gorance-teletrader
> 
> Thank for contributing. 😄
> I have left some comments on the PR.
> I also think this should be split into different PRs. (@saghul can you confirm 😄)

I susbribe to everything @akshitkrnagpal has said here :-) Plus an additional thank you for contributing!

gorance-teletrader:
Thanks for fast replay 👍 
i have fixed "problems" and ones that i have find while coding this

I don't have time to split PRs, but i agree that is better (i should think about it when i started)

app/features/config/index.js
```diff
@@ -28,5 +28,26 @@ export default {
...
      * The URL of Terms and Conditions Page.
      */
-    termsAndConditionsURL: 'https://jitsi.org/meet/terms'
+    termsAndConditionsURL: 'https://jitsi.org/meet/terms',
+
+    /**
+     * Some of default values for application.
+     */
+    defaults: {
+        windowAlwaysOnTop: true
```
akshitkrnagpal: I think `alwaysOnTopWindowEnabled` would describe this flag better. 😄

package-lock.json
```diff
@@ -5944,8 +5958,17 @@
       "requires": {
         "nan": "^2.2.1",
         "postis": "^2.2.0",
-        "prebuild-install": "^2.4.1",
-        "robotjs": "github:jitsi/robotjs#5cc469f655669e728b61271c2c57ce97d62976fd"
+        "prebuild-install": "^2.4.1"
+      },
+      "dependencies": {
+        "robotjs": {
```
akshitkrnagpal: I think the Travis build might be breaking because if this `package-lock.json` change. Since `package.json` does not have any change the update in lock file might be due to version mismatch of `npm` or `node`.

Travis uses node `v8.16.0` and `npm 6.4.1`.

Can you tell the version of node and npm you are using?

gorance-teletrader: Hi, i'm using 
Node: 10.15.3 
npm:   6.4.1 (same)

app/features/app/components/App.js
```diff
@@ -14,18 +16,58 @@ import { Welcome } from '../../welcome';
...
+     * Handler when main proccess contact us.
+     *
+     * @param {Object} event - Message event triggered by .
+     * @param {Object} arg - Object with two props same as Conference object.
+     *
+     * @returns {void}
+     */
+    _listenOnProtocolMessages = (event, arg) => {
+        // change route when we are notified
+        // eslint-disable-next-line no-invalid-this
```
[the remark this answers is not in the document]
gorance-teletrader: Sorry for this :)
I'm used to my react setup

app/features/config/index.js
```diff
@@ -28,5 +28,26 @@ export default {
...
+    /**
+     * Some of default values for application.
+     */
+    defaults: {
+        alwaysOnTopWindowEnabled: true
+    },
+
+    /**
+     * All needed configurations for storage in our application
+     * This is including electron-store and redux-persist
```
saghul: I don't want to have 2 ways to persist storage. Let's exist with the existing redux-persist please.

app/features/config/index.js
```diff
@@ -28,5 +28,26 @@ export default {
...
+     */
+    storage: {
```
saghul: This doesn't need to be configurable, please undo this change.

app/features/settings/components/WindowAlwaysOnTopToggle.js
```diff
@@ -0,0 +1,118 @@
...
+    /**
+     * Initializes a new {@code WindowAlwaysOnTopToggle} instance.
+     *
+     * @inheritdoc
+     */
+    constructor(props) {
+        super(props);
+
+        this.state = {
+            windowAlwaysOnTop: true
```
saghul: You don't need to store state here, just use the prop value.

app/features/config/index.js
gorance-teletrader: I don't understand comment, you need to use electron-store to access redux-persist data from main.js
And redux-persists uses electron-store to read/write data on user device

app/features/config/index.js
gorance-teletrader: Later, if you want to change **rootKey** in redux persists (change only in config) your app is still working, because main.js uses this config to access stored data

app/features/settings/components/WindowAlwaysOnTopToggle.js
gorance-teletrader: i will, but i was following style of [settings components](https://github.com/jitsi/jitsi-meet-electron/blob/master/app/features/settings/components/StartMutedToggles.js#L55)

akshitkrnagpal: @saghul 
I think this was done because of `isDefaultChecked` prop. Don't remember what problem was it causing.

main.js
```diff
@@ -141,23 +161,144 @@ function createJitsiMeetWindow() {
...
+/**
+ *
+ * Get data from store, and parse what is needed
+ * So other functions can access data as objects
+ *
+ * There is currently limitation
+ * persist data is typeof string
+ * so we cannot just do `store.get('a.b.c');
+ */
+function parsePersistStore() {
```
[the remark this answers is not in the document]
gorance-teletrader: Its my pleasure, i was thinking about that
but i need to find more free time to test this one

app/features/config/index.js
saghul: Oh, I'll check it again then.

app/features/config/index.js
saghul: We don't want that to be configurable.

app/features/settings/components/WindowAlwaysOnTopToggle.js
saghul: Oh, right!

main.js
```diff
@@ -141,23 +161,144 @@ function createJitsiMeetWindow() {
...
+    // This is required for some features
+    // Google Sign In - Start Youtube Live Stream, Settings -> Calendar
+    // Microsoft Sign In - Settings -> Calendar
+    // Dropbox Sign In - Start Recording Feature.
+    //
     initPopupsConfigurationMain(mainWindow);
-    setupAlwaysOnTopMain(mainWindow);
+
+    if (isWindowAlwaysOnTop) {
+        setupAlwaysOnTopMain(mainWindow);
```
akshitkrnagpal: Hey @gorance-teletrader 
Hey @gorance-teletrader 

I just took another look and I found that `setupAlwaysOnTopRender` is still being executed in Conference component. Can you make that prop dependent? 😄 

Also, please check what is the effect of keeping `setupAlwaysOnTopMain` and just not doing `setupAlwaysOnTopRender` if the prop it false. 

It might clean up main a lot. 😄

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "jitsi-meet-electron",
-  "version": "1.1.1",
+  "version": "1.2.0",
```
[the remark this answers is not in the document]
akshitkrnagpal: On the second thought, you can remove the version bump from this. We might not release a new version immediately after this PR.

app/features/config/index.js
wilhelmy: I assume this is about the visibility of the little video window... A friend of mine and I are having problems with that window popping up whereever we go, and we both find it kind of intrusive about using the electron client. I think he fixed it by some workaround that closes it whenever it appears, I just wanted to mention that this actually causes usability problems for some people.

kempu:
Thank you @gorance-teletrader for this awesome code, saved me days of research :)
To get this functionality released, I have updated the code to the latest app version and split these functionalities into separate pull requests (#263 and #264) based on maintainers request.

saghul:
Continues in #263 and #264

PR #174, opened 2019-07-09
Adds using power monitor and detecting computer suspend.

damencho:
Depends on https://github.com/jitsi/jitsi-meet-electron-utils/pull/48, new jitsi-meet-electron-utils release and bumping the package.json version.

### DIFF
--- a/app/features/conference/components/Conference.js
+++ b/app/features/conference/components/Conference.js
@@ -12,7 +12,8 @@ import {
     setupScreenSharingForWindow,
     setupAlwaysOnTopRender,
     initPopupsConfigurationRender,
-    setupWiFiStats
+    setupWiFiStats,
+    setupPowerMonitorRender
 } from 'jitsi-meet-electron-utils';
 
 import config from '../../config';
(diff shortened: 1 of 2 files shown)


main.js
```diff
@@ -12,7 +12,7 @@ const windowStateKeeper = require('electron-window-state');
 const {
     initPopupsConfigurationMain,
     getPopupTarget,
-    setupAlwaysOnTopMain
```
saghul: Is this no longer necessary?

damencho: That is a mistake, will fix it.

damencho: Fixed.

app/features/conference/components/Conference.js
```diff
@@ -265,10 +266,13 @@ class Conference extends Component<Props, State> {
         new RemoteControl(iframe); // eslint-disable-line no-new
         setupAlwaysOnTopRender(this._api);
         setupWiFiStats(iframe);
+        setupPowerMonitorRender(iframe);
 
+        this._api.on('suspendDetected', (event: Event) => {
```
hristoterezov: Maybe add `this. _onVideoConferenceEnded` as a value for the listener? You may need to do:`this. _onVideoConferenceEnded = this._onVideoConferenceEnded.bind(this)` in the constructor but I'm not sure.

PR #197, opened 2020-02-20
integrate screen sharing tracker window

andrei-gavrilescu:
This PR depends on https://github.com/jitsi/jitsi-meet-electron-utils/pull/66.

### DIFF
--- a/main.js
+++ b/main.js
@@ -13,7 +13,8 @@ const {
     initPopupsConfigurationMain,
     getPopupTarget,
     setupAlwaysOnTopMain,
-    setupPowerMonitorMain
+    setupPowerMonitorMain,
+    setupScreenSharingMain
 } = require('jitsi-meet-electron-utils');
 const path = require('path');
 const URL = require('url');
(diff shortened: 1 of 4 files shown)


package-lock.json
```diff
@@ -6052,21 +5976,13 @@
       }
     },
     "jitsi-meet-electron-utils": {
-      "version": "github:jitsi/jitsi-meet-electron-utils#3b6562f4d6104bfb5099414d1a79d4164767ea9b",
```
hristoterezov: I've just created a jitsi-meet-electron-utils release. Could you please rerun npm install?

main.js
```diff
@@ -149,6 +150,7 @@ function createJitsiMeetWindow() {
     initPopupsConfigurationMain(mainWindow);
     setupAlwaysOnTopMain(mainWindow);
     setupPowerMonitorMain(mainWindow);
+    setupScreenSharingMain(mainWindow, 'Jitsi Meet');
```
saghul: Please don't hardcode this. Use https://github.com/jitsi/jitsi-meet-electron/blob/3c448837e771796018dfff46dda2b89c7e5f98e2/app/features/config/index.js#L11

andrei-gavrilescu: ok, will create a PR to fix this.

saghul: Thanks!

PR #235, opened 2020-04-08
Fix chrome sandbox

saghul:


saghul:
For posterity. This has somehow not worked.

PR #261, opened 2020-04-11
Pr/172

kempu:
This excellent functionality is just laying around here.  Let's get it in order and merged :)
I took the pull request #172 and updated to the latest codebase, moved the app protocol prefix to the config file. Also made some amendments from code review request from the pull request #172.


saghul:
Thanks for bringing the up to date. Can you please split the PRs into 2? They are unrelated and having them intertwined will only make them stall again I’m afraid.

kempu:
Sure, that actually makes sense. I'll close this one, and send two new pull requests.

PR #263, opened 2020-04-12
Add protocol to open conference links with the app

kempu:
This commit will add application protocol to open links directly into the conference (issue #168).
This feature is based on the code from pull request #172 by @gorance-teletrader.


awlx:
We got some reports, that this breaks the ability to open a different server than the one configured in settings. Can you confirm?

app/features/utils/functions.js
```diff
@@ -54,3 +54,36 @@ export function normalizeServerURL(url: string) {
...
+        room = inputURL;
+    }
+
+    // Don't navigate if no room was specified.
+    if (!room) {
+        return;
+    }
+
+    return {
+        room
```
awlx, line 87: You only return the room here but not the server url. So it will default to the default serverURL.

saghul: Yep, please port this logic here: https://github.com/jitsi/jitsi-meet-electron/pull/263/files#diff-8265ab18888d94efa59fa6ff246f1688L128

app/preload/preload.js
```diff
@@ -8,6 +8,6 @@ window.jitsiNodeAPI = {
     createElectronStorage,
     osUserInfo: os.userInfo,
     shellOpenExternal: shell.openExternal,
-    jitsiMeetElectronUtils
+    jitsiMeetElectronUtils,
+    ipcRenderer
```
saghul, line 12: @andrei-gavrilescu Can you PTAL? I wonder if it's safe to do this.

app/features/utils/functions.js
kempu: Thanks for testing, Annika!
Is there any documentation about how it was expected to be working? I was unable to find any. 
I did see some comments mentioning server option in the app link, but this seemed like a security risk. It makes phishing attacks easier by redirecting users to call server they were not configured to use.
If this is the expected way to use the app link, then there should be some option for admins and users to protect themselves from potential phishing attempts.

saghul: Fail point. The applink could just take the room into account. Then it would ue the configured server in the app. The problem with your change is that hanndling of the server URL seems to be ccompletely gone. Users should still be able to use the configured server.

awlx: It caused lots of confusion for my users that the app behaviour changed to not being able to provide a different server than the configured one. So at least in the App input field the server should be treated as before.

app/preload/preload.js
andrei-gavrilescu: Hello everyone.
The electron recommendation in this case is to expose a custom API rather than directly expose node functionality to a renderer that loads remote content. 
Wrapping the ipcRenderer logic inside a function and exposing that would be the preferred way.
See: https://www.electronjs.org/docs/tutorial/security point 2.

saghul: That's what I was fearing. Thanks a lot @andrei-gavrilescu ! @kempu can you please adjust the code?

app/features/utils/functions.js
kempu: Thank you both for the feedback. Could you please provide example format/links how it's expected to be working. As the previously linked code sample was commented in code review as "not working", then I wouldn't trust that as an example.
PS! @saghul, I'm bit confused about your comment regards security. Could you please maybe clarify your thinking please, thanks! :)

app/preload/preload.js
kempu: Thanks! Will try to take a look at it as soon as I get some extra time.

app/features/utils/functions.js
awlx: Previously you could enter https:///somejitsiserver/room and it would take that as an input and also connect to that server.

Now it's also taking that input but strips the server name and connects straight to the configured one.

saghul: @kempu My comment was about the potential security implications of exposing that.

kempu: Thanks, everybody for the help. Hope the updated code suits better :)

app/preload/preload.js
kempu: Thanks for catching that. Hope the new approach is better :)

awlx:
I tested it in our fork. It seems to work well now :). Even with other servers.

chamerling:
Hello @saghul, it seems that the changes you requested has been addressed. Can you validate then merge?

daytonturner:
are there also changes required for https://meet.jit.si to attempt to launch the desktop client if its installed, using the protocol handler, or what will accomplish that?

chamerling:
@daytonturner you are right, there is nothing on the jitsi web application side for this. Anyway, I think that it is not the plan for now, and that this new protocol have to be displayed/used by external applications

daytonturner:
@chamerling I'll take a look into the jitsi-meet repo and see if i can leverage some of the work done for the mobile client to display the deep linking launch page.  There are some javascript libraries available that can detect if a protocol handler is registered and if so, make a different choice about what to do display for the user.

package.json
```diff
@@ -56,7 +56,16 @@
...
     "directories": {
       "buildResources": "resources"
-    }
+    },
+    "protocols": [
+      {
+        "name": "jitsi-protocol",
+        "role": "Viewer",
+        "schemes": [
+          "jitsi"
```
saghul, line 65: can we please make this jitsi-meet ?

kempu: I can see you already found the answer :) 
Sure thing, just pushed the change commit.

bossoq:
@kempu Nice work! Tested on macOS and Ubuntu, it's worked. However, on Windows, it's just start the application but not entering to the meeting room (Both meet.jit.si and self-hosted Jitsi Meet).

chamerling:
@kempu @bossoq Hello.

Same issues here. Works well on OS X but have issues on Linux and Windows which does not set the server URL correctly

saghul:
@kempu can you please take a look at what @bossoq and @chamerling report? Also, there is a small conflict now (sorry, my bad).

chamerling:
@kempu Hello,

Do you need some help on this subject? Having it working on Windows is something quite important for the community.

kempu:
Thanks, @chamerling!
Work is taking all my time right now, so I'm quite slow with side projects at the moment. But I added your kindly provided fix. 
As I don't have a Windows machine at hand right now, could please you test to make sure I got everything.

chamerling:
Thanks @kempu for this. Next time you should `cherry-pick` commit to keep author ;)
Going to make some tests at least on OS X and Windows first. Will ask my QA team to do more soon.
Anyway, @saghul what about being able to define deeplinks with Jitsi instance in it like `jitsi-meet://jitsi.company.com/room` ? The current code does not support this and I am pretty sure that the mobile app does.

kempu:
That’s a good point, will do a cherry pick next time 👍
Please add the room name in app link suggestion as a feature request or a new pull request if you have the code.
I committed a small change I did for our own copy of Jitsi, I had no intention to invest this much time into this.

chamerling:
> Anyway, @saghul what about being able to define deeplinks with Jitsi instance in it like `jitsi-meet://jitsi.company.com/room` ? The current code does not support this and I am pretty sure that the mobile app does

@saghul then what about this?
I suggest to merge as is and then I will send a PR to support this feature.

main.js
```diff
@@ -209,6 +218,45 @@ function createJitsiMeetWindow() {
...
         mainWindow.show();
     });
+
+    /**
+     * This is for windows [win32]
+     * so when someone tries to enter something like jitsi-meet://test
+     *  while app is closed
+     * it will trigger this event below
+     */
+    handleProtocolCall(process.argv[2]);
```
chamerling, line 228: @kempu this is bad. You must remove this line (a cherry-pick solved this...)

saghul:
> @saghul then what about this?
> I suggest to merge as is and then I will send a PR to support this feature.

Sorry I missed this, needed a break. I can cut a release when we're ready, so I'd rather have thing right from the get go, since it's not a big change anyway.

chamerling:
@saghul yes it will be better to have everything OK, but since I am not this PR creator, it is quite complex to collaborate on it. Can't we change the target branch to a feature one, work on it and then merge it into master when done.
The change I want to send is just some lines and takes so much time currently

kempu:
I would recommend to have a wider open discussion about this feature. I did not add it intentionally, as it opens up a phishing vulnerability. 
For example, if an organisation is using Jitsi, someone can pretend to be you and send a call link to a different server while still using the company configured desktop app that the user trusts. It’s just too common attack vector to just ignore it. 
If this feature really is that important for some, then I would at least recommend adding a config option for admins to disable it or a big warning dialog that warns user they are about to connect to an unknown server.

chamerling:
This is already how the mobile app works so I do not see any issue here. I do not think we are going to ask the Jitsi guys to change how the mobile app works. Do we?

kempu:
Imho, if a problem exists in one system does not mean it should be replicated blindly to new systems just because “this is how it’s done previously”. Instead it would be wise to take a step back for a moment and open up a discussion about it 🙂
But as I’m not involved in this project so heavily, and I have raised my concerns about the security implications, then I’ll leave it up to you more active participants to decide how to run your project 🙂

chamerling:
Well I understand your point. This is a global discussion but IMO there are different deployments and different needs. Jitsi tends to be quite open on this purpose and so being able to open conferences on different servers follows this.
Anyway if you are inviting someone from another company, you can not force him to configure the client before clicking the link.  Most users are standard ones and evzn do not understand how to change a setting. So clicking a link should do everything. This is up to the link provider to display a warning message at the end, not to the app. Don't you think?

awlx:
I would also say it's easier for a user to have a consistent behaviour across platforms. Also yes, it technically adds a risk. 

Maybe add a warning hint if server!=configuredserver?

kempu:
Yes, @awlx, exactly 👍 That’s what I meant that at least a warning should be displayed for the user if servers do not match. 
And perhaps there still could have an option in the config file to disable it. Even if it’s enabled by default, then at least it gives an option for organisations to disable it on their own deployments without the need to change code.
Re @chamerling, leaving the responsibility of warning to the link provider would not work. As in the case of phishing, the link provider is the attacker 🙂

chamerling:
#389 is merged and replaces this one, you can close this PR now.

kmsrao:
With the compiled code, I am still not able to open the App from the link. The App opens but does not join the room. Is this fixed ? or still open ?

I am using Windows 10.

chamerling:
Hello @kmsrao

The link must be like `jitsi-meet://room` or `jitsi-meet://server:port/room`.
Which one did you use?

PR #286, opened 2020-04-20
Add support for query parameters

chamerling:
Hello everybody,

This PR adds the support for query parameters in the Jitsi URL in order to fill extra fields in `options` when calling `api = new JitsiMeetExternalAPI(domain, options)`.

The initial usecase was to be able to open conference with a JWT token: The `options.jwt` parameter is set from the `?jwt=` query parameter. This usecase has been validated, you can open a room protected with jwt with an input URL like `https://HOST:PORT/ROOM?jwt=TOKEN`.
This is compliant with browser-based JWT protected rooms where the `jwt` token is also defined as query parameter.

The approach is quite generic and allows simple parameters in `options` to be defined from the user. While there is nothing more than a simple input field available to create a room, this is the only approach which allows to pass other parameters. Anyway, if you think that it is too much generic, I am OK to change the code to support only a set of query parameters instead of overriding all the options with input ones.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,8 @@ npm install
 npm start
 ```
 
+The debugger tools are available when running in dev mode and can be activated with keyboard shortcuts as defined here https://github.com/sindresorhus/electron-debug#features.
+
 #### Building the production distribution
 
 ```bash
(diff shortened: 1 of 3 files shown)


MichaelBailly:
Thank you for that feature @chamerling ! I hope it'll get merged

chamerling:
Hello Jitsi team,

Please tell me if I need to rework this PR.

Thanks a lot

chamerling:
Hello. Any chance to get this PR reviewed?
Thanks a lot

saghul:
Sorry for the delay. Thanks for thr fixes!

PR #289, opened 2020-04-21
macOS Catalina Fix for Camera/Microphone

imdatsolak:
Without the "extendInfo" and "hardenedRuntime=false", the OS doesn't even ask for permission for camera & microphone when the client is started first time on macOS Catalina. The "extendInfo" provides the messages describing why we need access to camera/microphone.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,12 @@
     "mac": {
       "artifactName": "jitsi-meet.${ext}",
       "category": "public.app-category.video",
-      "darkModeSupport": true
+      "darkModeSupport": true,
+      "hardenedRuntime": false,
+      "extendInfo": {
+        "NSCameUsageDescription": "jitsi meet requires access to your camera in order to make video-calls.",
+        "NSMicrophoneUsageDescription": "jitsi meet requires access to your microphone in order to make calls (audio/video)."
+      }
     },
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -28,7 +28,12 @@
     "mac": {
       "artifactName": "jitsi-meet.${ext}",
       "category": "public.app-category.video",
-      "darkModeSupport": true
+      "darkModeSupport": true,
+      "hardenedRuntime": false,
+      "extendInfo": {
+        "NSCameUsageDescription": "jitsi meet requires access to your camera in order to make video-calls.",
+        "NSMicrophoneUsageDescription": "jitsi meet requires access to your microphone in order to make calls (audio/video)."
```
wigging, line 35: Camera in the key is misspelled, it should be `NSCameraUsageDescription`. Also, `jitsi meet` should be capitalized.

```
"NSCameraUsageDescription": "Jitsi Meet requires access to your camera in order to make video-calls.",
"NSMicrophoneUsageDescription": "Jitsi Meet requires access to your microphone in order to make calls (audio/video)."
```

imdatsolak: Thanks, has been fixed in the original, will close and send a new request

PR #296, opened 2020-04-23
Add timeout configuration in settings

chamerling:
Hello,

On some jitsi deployments, the current 10 seconds timeout allowed before displaying error is a little too small. The current PR adds a configuration field in settings to be able to change this value as shown in the screenshots below.

Capture d’écran 2020-04-23 à 17 10 58
Capture d’écran 2020-04-23 à 17 10 46

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,6 +85,7 @@
     "@atlaskit/navigation": "33.3.8",
     "@atlaskit/onboarding": "6.1.14",
     "@atlaskit/page": "8.0.12",
+    "@atlaskit/panel": "0.3.5",
     "@atlaskit/spinner": "9.0.13",
     "@atlaskit/theme": "7.0.1",
     "@atlaskit/toggle": "5.0.14",
--- a/app/features/conference/components/Conference.js
+++ b/app/features/conference/components/Conference.js
@@ -52,6 +52,11 @@ type Props = {
     _serverURL: string;
 
     /**
+     * Default Jitsi Server Timeout.
+     */
+    _serverTimeout: number;
+
+    /**
      * Start with Audio Muted.
      */
     _startWithAudioMuted: boolean;
@@ -120,6 +125,7 @@ class Conference extends Component<Props, State> {
     componentDidMount() {
         const parentNode = this._ref.current;
         const room = this.props.location.state.room;
+        const serverTimeout = this.props._serverTimeout || config.defaultServerTimeout;
         const serverURL = this.props.location.state.serverURL
             || this.props._serverURL
             || config.defaultServerURL;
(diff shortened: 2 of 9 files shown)


chamerling:
Hello. Any chance to get this PR reviewed?
Thanks a lot

awlx:
I personally would hide this in an expert menu. Normal users are already flashed by the options ... and adding a timeout value might confuse them even more :/.

saghul:
Not sure I'd have this option. 10 seconds is a very conservative value. What kind of video quality can one expect from a server which takes 10 seconds to reply to a simple request fetching a config file?

saghul:
> I personally would hide this in an expert menu.

Agreed, if we have some "advanced options" this could go in there.

chamerling:
> which takes 10 seconds to reply to a simple request fetching a config file?

So from the code this is not what is this timeout used for. The timeout is used from the launch of the external Jitsi client (ie the iFrame with all the jitsi assets) to the conference joined event.

chamerling:
> Agreed, if we have some "advanced options" this could go in there.

But we do not have. So what do you suggest?

awlx:
Create an advanced menu? :)

saghul:
Maybe we could have a collapsible "Advanced" section in the settings. We could use this for it: https://atlaskit.atlassian.com/packages/bitbucket/panel

lukas2511:
I think this should be a server-side option. This sounds like something where in a normal situation connectivity would be hardly usable anyway, but this would be extremely useful for something like server-side connection rate limiting, especially after a prosody crash where maybe thousands of reconnects would occur at the same time. That way they'd all need some more time for connecting, but nothing would error out and the server requests can be stretched over time.

chamerling:
> Maybe we could have a collapsible "Advanced" section in the settings. We could use this for it: https://atlaskit.atlassian.com/packages/bitbucket/panel

OK, will try to push this today

chamerling:
@saghul  Sorry for the delay, I pushed some changes to have the advanced settings panel like this:

Capture d’écran 2020-05-13 à 13 01 33
Capture d’écran 2020-05-13 à 13 01 28

awlx:
That looks great!

package.json
```diff
@@ -85,6 +85,7 @@
     "@atlaskit/navigation": "33.3.8",
     "@atlaskit/onboarding": "6.1.14",
     "@atlaskit/page": "8.0.12",
+    "@atlaskit/panel": "^0.3.5",
```
saghul, line 88: Can you please pin the depenency to the exact version?

chamerling: Fixed sorry for this.

PR #308, opened 2020-04-26
Catalina fix

imdatsolak:
Fixed an issue with macOS Catalina requiring additional meta information before allowing Camera/Microphone usage. Also, the "hardenedRuntime" needs to be disabled, otherwise it will require notarization.


saghul:
Cheers!

PR #317, opened 2020-04-30
Fix shortcuts on macOS

dbkr:
The regular windows shortcuts (eg. minimise, hide) don't work unless
they have menu entries, so add them along with the standard set of
expected menu items for mac (although I've omitted 'about' since it
opens a window so is a slightly more visible change).

Also Electron has a role for 'quit' so use this instead of providing
our own.


saghul:
Thank you!

PR #342, opened 2020-05-20
Refactor avatar, add random room name generator

saghul:
Screen Shot 2020-05-20 at 09 50 38


saghul:
> The 'GO' button doesn't seem to have any separation from the conference input box, looks a bit weird but that might be just me.

Agreed. My CSS fu ran out. If you (or anyone) knows how to add a bit of separation that would be ideal indeed.

PR #343, opened 2020-05-20
Update to Electron 9 to support E2EE

saghul:
Electron 9 ships with Chrome 83 which means we have access to E2EE by
enabling experimental web platform features.

Screen Shot 2020-05-20 at 10 05 00

### DIFF
--- a/main.js
+++ b/main.js
@@ -23,6 +23,9 @@ const config = require('./app/features/config');
 // We need this because of https://github.com/electron/electron/issues/18214
 app.commandLine.appendSwitch('disable-site-isolation-trials');
 
+// Needed until robot.js is fixed: https://github.com/octalmage/robotjs/issues/580
+app.allowRendererProcessReuse = false;
+
 autoUpdater.logger = require('electron-log');
 autoUpdater.logger.transports.file.level = 'info';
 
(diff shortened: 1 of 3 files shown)


main.js
```diff
@@ -153,6 +156,7 @@ function createJitsiMeetWindow() {
         minHeight: 600,
         show: false,
         webPreferences: {
+            experimentalFeatures: true, // Insertable streams, for E2EE.
```
fippo, line 159: can we use the more specific RTCInsertableStreams field trial? the command line flag for that is --force-fieldtrials=RTCInsertableStreams/Enabled

saghul: Never tried that on Electron. Any reason why this might be undesirable?

fippo: not sure what other experimental features we enable by this. OTOH we don't use them...

saghul: this is equivalent to "experimental web platform features" I'll check the field trial thing.

saghul: Looks like that one is not supported: https://www.electronjs.org/docs/api/command-line-switches

Nokia808:
@saghul 
Dear from my information is that Jitsi is already end-to-end encrypted as long as only 2 persons in the video conference room. But when they become 3 or more, then data will decrypted at the level of server for need of video bridge to work ..... So, what this PR will added if it is merged ?

saghul:
@Nokia808 See here: https://jitsi.org/blog/e2ee/

Nokia808:
@saghul 
Really amazing ! I'm & my students really thankful for your very hard efforts !

Issue #299, opened 2020-04-24
Camera access issue


Triskele69:
I have an error Failed to access your camera for an unknown reason. Sometimes error about no permissions.
Checked in ProcessManager that no other programs are using my camera.
Sharing desktop is working.
And everything is working in browser.
Connection to server is https with valid cert.

Cameras not working on every computer. Win7 and Win10.

What could it be?

iarani:
Facing the same issue on macOS Mojave version 10.14.6. Can access it during development but not after creating a distribution file. Tried manually editing info.plist to update <key>Privacy - Camera Usage Description</key> to a more "proper" description. Camera turns on for a second and then shows the error "Your camera is in use by another application"

iarani:
Resolved it by adding askForMediaAccess("camera") explicitly on Mac 

`const { systemPreferences } = require('electron');`
`if(systemPreferences.askForMediaAccess("camera")) { createJitsiMeetWindow(); } `

awlx:
Where did you patch this @iarani and does it also work for other OS?

saghul:
Any chance you were using an old version of the code?

awlx:
The issue seems to persists with the new version :/.

saghul:
I’m running Mojave and haven’t run into this whole testing the release. Any specific steps I should follow to try to repro?

awlx:
I am still trying to repro this in a good way ... It seems to only fail for some users which makes it hard to track down -.-.

Echolon:
Hi,
If you have an issue with your camera, please see the pinned issue #5425, which has more information on how to help the Jitsi team to resolve the issue.

---
*I'm a community contributor and not directly affiliated with the Jitsi team.*

saghul:
We have made many changes to handle these in Jitsi Meet. Please test the latest version, using meet.jit.si as the server.

PR #351, opened 2020-05-22
Fix Camera permissions for signed Apps

awlx:
This fixes the camera permissions if the app needs to be signed.

Should also fix:
#299

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,9 @@
       "artifactName": "jitsi-meet.${ext}",
       "category": "public.app-category.video",
       "darkModeSupport": true,
-      "hardenedRuntime": false,
+      "hardenedRuntime": true,
+      "entitlements": "entitlements.mac.plist",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -29,7 +29,9 @@
       "artifactName": "jitsi-meet.${ext}",
       "category": "public.app-category.video",
       "darkModeSupport": true,
-      "hardenedRuntime": false,
+      "hardenedRuntime": true,
```
saghul, line 32: Does this also work in the non-signed case?

awlx: It should not, as the entitlements also get used to ask for permission.

awlx: https://developer.apple.com/documentation/security/hardened_runtime?preferredLanguage=occ

saghul:
Verified it works on unsigned apps too, thanks a lot @awlx !

PR #381, opened 2020-06-03
Add SHOW_DEV_TOOLS environment variable

chamerling:
This PR allows to open devtools from the command line when launching the application such as:

```sh
SHOW_DEV_TOOLS=true npm start
```

It also adds the command line support when launching app (https://www.electronjs.org/docs/api/command-line) with `--show-dev-tools`

### DIFF
--- a/main.js
+++ b/main.js
@@ -21,7 +21,7 @@ const path = require('path');
 const URL = require('url');
 const config = require('./app/features/config');
 
-const showDevTools = Boolean(process.env.SHOW_DEV_TOOLS);
+const showDevTools = Boolean(process.env.SHOW_DEV_TOOLS) || (process.argv.indexOf('--show-dev-tools') > -1);
 
 // We need this because of https://github.com/electron/electron/issues/18214
 app.commandLine.appendSwitch('disable-site-isolation-trials');
--- a/main.js
+++ b/main.js
@@ -21,6 +21,8 @@ const path = require('path');
 const URL = require('url');
 const config = require('./app/features/config');
 
+const showDevTools = Boolean(process.env.SHOW_DEV_TOOLS);
+
 // We need this because of https://github.com/electron/electron/issues/18214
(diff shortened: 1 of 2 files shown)


main.js
```diff
@@ -21,6 +21,8 @@ const path = require('path');
 const URL = require('url');
 const config = require('./app/features/config');
 
+const showDevTools = Boolean(process.env.SHOW_DEV_TOOLS);
```
saghul, line 24: How about doing a command line argument too?

package-lock.json
```diff
@@ -4,106 +4,100 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
-    "7zip-bin": {
```
saghul, line 7: What's with this entire diff? You haven't changed package.json even :-O

main.js
saghul: That would also work on release builds for example.

chamerling: Good idea, adding it

package-lock.json
chamerling: Umh, looks like my `npm update` changed things... Removing this file from the PR

main.js
chamerling: Add command line support

package-lock.json
chamerling: Removed

main.js
```diff
@@ -21,6 +21,8 @@ const path = require('path');
 const URL = require('url');
 const config = require('./app/features/config');
 
+const showDevTools = Boolean(process.env.SHOW_DEV_TOOLS) || app.commandLine.hasSwitch('show-dev-tools');
```
saghul, line 24: Wait, this is not going to work. Doesn't `app.commandLine`refer to the Chromium command line args?

chamerling: @saghul true, sorry for that. I updated the PR

saghul:
Cheers!

Issue #379, opened 2020-06-03
POLL: Should the avatar go?


saghul:
Since this repo has a small (but enthusiastic) group of individuals, let's try this.

There is a disconnect between the avatar displayed in the app and what the web would use, should you use a browser. Since we are already asking for a username and email, the embedded jitsi meet will already know to use gravatar or the locally generated avatar, or whatever is configured.

There is no good way to embed the avatar logic in Jitsi Meet here.

So I'm thinking of getting rid of it from the UI here. Thoughts? Please vote with 👍  or 👎  and leave a comment if you need to elaborate.

dchmelik:
Just allow people to open their own: don't force gravatar.

PR #385, opened 2020-06-04
Remove local avatar

saghul:
The deployment will generate the right one based on the provided name
and email.

Fixes: https://github.com/jitsi/jitsi-meet-electron/issues/379


dchmelik:
Re-add local avatar.  Many people won't use ones from email for security/privacy (some of theirs, and email addresses' & gravatar avatars are blank for that reason , but I want my face in jitsi).

PR #391, opened 2020-06-11
Bundle external api

saghul:


saghul:
> but how can we be sure that the bundled one is valid for the selected jitsi server?

We use a very very small portion of it, which hasn't changed in years, so there is that :-)

bgrozev:
Do we have to bundle a minified version?

chamerling:
@saghul Hello,

Coming back to this subject: There are still references to the server external api JS file in the code and in the README. Now that it is bundled with the app, don't you think this must be cleaned?

Issue #334, opened 2020-05-14
[Feature request] Add option to delete program catch.


Nokia808:
Hi. Currently it is not possible to delete program catch unless manually by delete:
home/myusername/.config/Jitsi Meet

It will be better if you add an option to "Delete program catch" in the "setting" to achieve this.

saghul:
Fixed in https://github.com/jitsi/jitsi-meet-electron/pull/392

Nokia808:
@saghul 
Hi dear. Thank you very much for your very hard efforts to make this application better & better !
But, only one thing: does removal of conference(s) is equal to removal of cache ? No, it is not equal ! In fact, it is a removal of part of cache not all cache ... !
What about other element, like custom instance setting ? I asked about an option to remove ALL directories that created by run the AppImage ... This what I meant by removal the cache.

Kindly, if you review the state of this issue ...

Issue #322, opened 2020-05-02
[Feature request] Allow user to delete a room from heir/his UI


Nokia808:
Hi dears.

After trying this nice package, I found an issue need to be covered:
Suppose that a user created a conference room. Then after, she/he found that it's name not appropriate or not long enough, so she/he will create new room with suitable name. Now, user not like to use previous room any more ... But the old room that user has no need to use it more will stick in the UI of application & can not removed currently ... 

I suggest on you to add new option that allow user to delete already created/joined room(s) from application UI.

Please notice that my request is allow user to delete already created/joined foom(s) from application UI not from server ... I'm not talking about server, but I'm talking about local deletion from UI ...

I think this new feature should be available on your official web application & all other platform specific applications ....

svoop:
Second that, it's really cumbersome and error prone to dive into the JSON file and remove the right parts by hand to get rid of a room on the UI.

Nokia808:
Currently I'm enforced to manually delete

home/myusername/.config/Jitsi Meet

folder so as to delete an unused room URL but this will delete all program catch ! This is bothering ...

saghul:
A PR would be welcome.

svoop:
@saghul It will be a while before I can look into that, got a shoutload of work for other FOSS projects on my desk already. So maybe someone else wants to chime in? Shouldn't be too difficult.

Echolon:
I think this is a duplicate of #127 and #283

Issue #283, opened 2020-04-17
Suggestion: Remove items from "recent list"


asmaloney:
(Just saw #127 which asks for a way to clear the list - this suggestion is similar but gives more control to the user.)

It would be useful to be able to remove items from the recent list OR provide a way to save a list of favourite rooms and show them at the top (this would be preferred of course, but sounds like more work).

For those of us that only use this with a few people (not using it for hundreds of different chats), it would be nice to be able to remove rooms we won't use in the future. (e.g. I have a few "test" rooms and one-offs listed.)

It could be a close "x" in the upper-right of each room box, or a right-click context menu to remove it.

(I would tackle this myself, but I'm a C++ developer and have not been initiated into the ways of the electron 😀.)

Echolon:
@asmaloney Isn't this a duplicate? or would make more sense to add you suggestions to the previous issue?

saghul:
Fixed in https://github.com/jitsi/jitsi-meet-electron/pull/392

Issue #127, opened 2018-09-29
[FR] Clear chat history or at lest remove the old chat rooms


Ark74:
This feature request is in the sense that the app saves the complete history of chats.
So it would be nice to be able to clear it or remove old chat rooms from the list in order to keep a slim list.

![ft-clear-list](https://user-images.githubusercontent.com/7444518/46251238-0c9f7680-c413-11e8-9dd1-55bd74fe7a2c.png)

Cheers!

saghul:
Makes sense, thanks for the feedback!

demlak:
Is it possible to clear cache or something like this to archive deleting rooms?

janwels89:
this seems to be stored on the client-side in your browser cookies. Try loading the page with an incognito tap to confirm.

Just delete your local browser cookies and the list should be empty again.

saghul:
Fixed in https://github.com/jitsi/jitsi-meet-electron/pull/392

PR #392, opened 2020-06-12
Fixes #313, #283, #322, #127, #334 - Add button to remove conference from recents list

### LINKED ISSUE #313: Deleting or bookmarking old session
When we launch jitsi-meet-electron, we can see our last "conference room". 
Is there a way to remove some or all of them?  Is there a way to create favorite ones?

freddytuxworth:
This PR makes it possible to remove conferences from the recents list, for convenience/privacy.
There are a bunch of issues about it: #313, #283, #322, #127, #334.

### DIFF
--- a/app/features/recent-list/components/RecentList.js
+++ b/app/features/recent-list/components/RecentList.js
@@ -6,6 +6,7 @@ import { connect } from 'react-redux';
 import type { Dispatch } from 'redux';
 import { push } from 'react-router-redux';
 
+import { conferenceRemoved } from '../actions';
 import {
     ConferenceCard,
     ConferenceTitle,
@@ -13,6 +14,8 @@ import {
     TruncatedText
 } from '../styled';
 import type { RecentListItem } from '../types';
+import Button from '@atlaskit/button';
+import CrossIcon from '@atlaskit/icon/glyph/cross';
 
 type Props = {
 
@@ -59,6 +62,20 @@ class RecentList extends Component<Props, *> {
     }
 
     /**
+     * Creates a handler for removing a conference from the recents list.
+     *
+     * @param {RecentListItem} conference - Conference Details.
+     * @returns {void}
+     */
+    _onRemoveConference(conference: RecentListItem) {
+        return e => {
+            this.props.dispatch(conferenceRemoved(conference));
+            e.stopPropagation();
+        };
+    }
+
+
+    /**
      * Renders the conference card.
      *
      * @param {RecentListItem} conference - Conference Details.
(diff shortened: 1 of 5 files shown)


awlx:
Nice! This will make many people very happy! :)

krombel:
Isn't it more wise to implement it in https://github.com/jitsi/jitsi-meet? Or just "backport" it there so it is available for especially the web people (which are as well complaining about this missing feature).
[This](https://github.com/jitsi/jitsi-meet/blob/stable/jitsi-meet_4627/react/features/recent-list/components/RecentList.web.js) looks at the first look at a good starting point.
Any plans/insights on that?

freddytuxworth:
> Isn't it more wise to implement it in https://github.com/jitsi/jitsi-meet? Or just "backport" it there so it is available for especially the web people (which are as well complaining about this missing feature).
> [This](https://github.com/jitsi/jitsi-meet/blob/stable/jitsi-meet_4627/react/features/recent-list/components/RecentList.web.js) looks at the first look at a good starting point.
> Any plans/insights on that?

As far as I can see those are two completely separate recent lists, so it would need to be implemented separately in both places. Happy to do that though, I'll give it a go :)

awlx:
They recent list in the electron-client is kept separately I guess, because it also keeps track of different servers which die web UI can't do.

awlx:
The feature is working really well :). 

I merged it our fork for testing! It makes many users very happy!
Screenshot 2020-06-13 at 11 14 35

app/features/recent-list/reducer.js
```diff
@@ -65,6 +72,21 @@ function _insertConference(
...
+ * @param {Array<RecentListItem>} recentList - Previous recent list array.
+ * @param {RecentListItem} toRemove - Conference to be removed.
+ * @returns {Array<RecentListItem>} - Updated recent list array.
+ */
+function _removeConference(
+        recentList: Array<RecentListItem>,
+        toRemove: RecentListItem
+): Array<RecentListItem> {
+    return recentList.filter(
+        (conference: RecentListItem) => conference.room !== toRemove.room);
```
saghul, line 87: You also need to check the serverURL field, or you'd end up removing all conferences with the same name, no matter the server.

But actually, why not remove all elements which are not `toRemove` ?

app/features/recent-list/components/RecentList.js
```diff
@@ -5,14 +5,16 @@ import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import type { Dispatch } from 'redux';
 import { push } from 'react-router-redux';
-
```
saghul, line 8: please leave this newline

app/features/recent-list/reducer.js
freddytuxworth: Good point, not sure why I added the `.room`s

freddytuxworth:
Should be fixed :+1:

benczur45:
Hello there!

Is this available on Browser version too? 

Thanks in advance.

saghul:
No, that is a separate implementation.

luistriana28:
@saghul 

Is there gonna be an update for web browser?

and for dockerized one too?

saghul:
Removing items for the recent list on the web interface just landed in master today. It will be part of the next stable release.

PR #399, opened 2020-06-26
Add german and italian

awlx:


freddytuxworth:
@awlx I added a translation `settings.serverTimeout` after review of my PR that I had missed on the first pass, so they are missing from your translation fields. I think that is the only one I added but it might be worth checking.

awlx:
Ah ... yeah. Sorry :). Will fix it!

awlx:
It's fixed now @freddytuxworth :).

PR #405, opened 2020-06-29
Increase default timeout to 30 seconds

chamerling:
Hello,

We have many small deployments where the default timeout to 10 seconds is not enough at all.
30 seconds will be more than enough and will avoid having unexpected errors. BTW, the error is not clear, going to create a PR so it is more explicit.

### DIFF
--- a/app/features/config/index.js
+++ b/app/features/config/index.js
@@ -29,7 +29,7 @@ export default {
     /**
      * The default server Timeout in seconds.
      */
-    defaultServerTimeout: 10,
+    defaultServerTimeout: 30,
 
     /**
      * URL to send feedback.


saghul:
How is it possible that the iframe takes > 10 seconds to load?!

chamerling:
@saghul it is. I saw it several times.
Anyway, this is not only the iframe load but the right event listener to be called (`onConferenceJoined` or something similar), which can take more than 10 seconds.
It depends on how Jitsi is deployed and how/from where you connect to it.

chamerling:
Oh I am wrong on the `onConferenceJoined`. (This is not in this project but it probably needs to be handled here also... :) )
Anyway, I encountered timeouts "just for" iframe load with 10 seconds timeout on small deployments (yes they exists)

saghul:
I'm merging this tentatively.

Issue #278, opened 2020-04-14
Internationalization / i18n support


ctr49:
It would be great if the app had support for other languages than English.
i18next looks very promising.

pioneer:
I vote for i18n, should be relatively fast to translate once there is a facility for that.

luixxiul:
I would expect some progress would be made after translate.jitsi.org would eventually be back.

https://github.com/jitsi/jitsi-meet/pull/6551#pullrequestreview-404638316

> We are working to put back translate.jitsi.org back in place ... Just more priority tasks moves that back in the list. Will see how it goes with the weekly plan and can merge of we cannot finish the move in the following few weeks.

saghul:
> I would expect some progress would be made after translate.jitsi.org would eventually be back.

Not really because the few strings the app uses don't use i18n. If someone wants to get that effort started, I'll happily review, but I cannot commit to making it happen.

luixxiul:
Yeah, please tackle stuff on priorities @saghul 👍

chamerling:
Hello,

Is there someone working on this right now? If not I will propably open a PR.

Thx

saghul:
Not that I know of, thanks for volunteering!

freddytuxworth:
> Hello,
> 
> Is there someone working on this right now? If not I will propably open a PR.
> 
> Thx

@chamerling I had started a bit of work on adding i18n - have you made significant progress already?

chamerling:
@freddytuxworth I did not have time to do anything so please do.

PR #407, opened 2020-06-29
Open the conference with the app locale

chamerling:
As defined by https://github.com/i18next/react-i18next, we can open the conference with a defined locale. 
The work from #278 allowed to use the OS locale on the React application on the Desktop side, but did not propagate it on the Jitsi External API. This PR corrects this: OS locale is now used on both sides.

### DIFF
--- a/app/features/conference/components/Conference.js
+++ b/app/features/conference/components/Conference.js
@@ -7,6 +7,7 @@ import type { Dispatch } from 'redux';
 import { connect } from 'react-redux';
 import { push } from 'react-router-redux';
 
+import i18n from '../../../i18n';
 import config from '../../config';
 import { getSetting, setEmail, setName } from '../../settings';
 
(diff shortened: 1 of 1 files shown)


app/features/conference/components/Conference.js
```diff
@@ -201,7 +202,11 @@ class Conference extends Component<Props, State> {
         const roomName = url.pathname.split('/').pop();
         const host = this._conference.serverURL.replace(/https?:\/\//, '');
         const searchParameters = Object.fromEntries(url.searchParams);
-        const urlParameters = Object.keys(searchParameters).length ? searchParameters : {};
+        const locale = { lng: i18n.language };
+        const urlParameters = Object.keys(searchParameters).length ? {
```
saghul, line 206: Please drop the check from Object.keys, Object.fromEntries will always return an object, so we can destructure it even if empty.

app/features/conference/components/Conference.js
```diff
@@ -9,6 +9,7 @@ import { push } from 'react-router-redux';
 
 import config from '../../config';
 import { getSetting, setEmail, setName } from '../../settings';
+import i18n from '../../../i18n';
```
saghul, line 12: please sort this alphabetically, by import path, with deeper paths going first

app/features/conference/components/Conference.js
chamerling: True, sorry for that

app/features/conference/components/Conference.js
chamerling: Fixed

PR #408, opened 2020-06-30
Add Russian localization

KovalevArtem:


saghul:
There is a syntax error in the JSON file, please check the details in the CI consoles.

chamerling:
We may not need 3 commits for this. Can you squash them?

KovalevArtem:
I don't really understand what I have to do. 😔

PR #413, opened 2020-07-03
Fix join via protocol link on Linux

### LINKED ISSUE #411: join via protocol link not fully working on Linux
Jitsi Meet Electron release: 2.3.0
Platform: Linux (CentOS 8/GNOME/Xorg)

We just tried the new "join conference via protocol link" (#389) feature. We noticed that on Windows 10 it works perfectly out-of-the-box, but unfortunately not on Linux.

(Let's ignore for the moment the problem of the actual registration of a scheme handler in the desktop environment, which we had to do manually in order to get the client launched when clicking on jitsi-meet: URIs.)

The problem we observed is that if JME is not currently running, the invocation of "~/jitsi-meet-x86_64.AppImage jitsi-meet://meetingname" only results in the client being started, but then remaining on the welcome page where I can enter the meeting name manually.

If, however, the JME client is already running in this state, executing  "~/jitsi-meet-x86_64.AppImage jitsi-meet://meetingname" results in the already-running instance to immediately join the meeting.

We would have expected that the client join the meeting on the first invocation, just like it happens with the Windows client.

chref:
Perform command-line protocol argument handling on all platforms, not only
win32. This allows the client to join a meeting passed on the command-line
even if no instance of the client is already running.

Resolves #411

### DIFF
--- a/main.js
+++ b/main.js
@@ -222,14 +222,11 @@ function createJitsiMeetWindow() {
     });
 
     /**
-     * This is for windows [win32]
-     * so when someone tries to enter something like jitsi-meet://test
+     * When someone tries to enter something like jitsi-meet://test
      *  while app is closed
      * it will trigger this event below
      */
-    if (process.platform === 'win32') {
-        handleProtocolCall(process.argv.pop());
-    }
+    handleProtocolCall(process.argv.pop());
 }
 
 /**


chamerling:
I will test it on OS X asap

saghul:
Thanks!

Issue #414, opened 2020-07-03
Translation missing


jrie:
The string "Invalid Timeout" is not by default translated.

Please see here:
https://github.com/jitsi/jitsi-meet-electron/blob/e4742dbc995a10b5a1eef84ca65ced725cbd95a7/app/features/settings/components/ServerTimeoutField.js#L80

saghul:
Nice catch! Any chance you can provide a fix?

saghul:
Fixed in https://github.com/jitsi/jitsi-meet-electron/pull/415

PR #415, opened 2020-07-03
Adding translation for server timeout, #414 ..

jrie:
in German and English translations.


jrie:
Please note, other translations are missing this translation at the moment!

PR #428, opened 2020-08-03
Update README.md

ovari:


saghul:
What does this address? Not a fan of just changing the quote types just because.

ovari:
`64bits` should be [64-bit](https://en.wikipedia.org/wiki/64-bit_computing)

Issue #426, opened 2020-08-01
Flathub listing


ovari:
Can Jitsi Meet be released as a flatpak on [Flathub](https://flathub.org/)?

Thank you

ovari:
https://github.com/jitsi/jitsi-meet-electron/issues/231#issuecomment-613965940

saghul:
No, sorry.

We are limited in the resources we can spend on this project, and adding another target is not on the table.

Issue #320, opened 2020-05-01
[Packaging] Make "Jitsi Meet Electron" available as official flatpak package on FlatHub.


Nokia808:
Hellow.

I would like to suggest on you to make "Jitsi Meet" officially available by you as official flatpak package on FlatHub.

Flatpak is a universal Linux package that compatible with almost all Linux distros ...... By making Jitsi Meet available as flatpak on FlatHub, it will be available for almost all Linux distros & users.

Flatpak is secure package system with default sandbox. It offer users automatic updates, & being independent from system version.

FlatHub home page is:
https://www.flathub.org/home
FlatHub repositories are here:
https://github.com/flathub
To post your flatpak on FlatHub, you have to make PR on:
https://github.com/flathub/flathub/pulls

Flatpak documentation here:
https://docs.flatpak.org/en/latest/

I forgot to say that flatpak is easier to be packaged than AppImage & flatpaks on FlatHub all are signed & the administrations of FlatHub prefer for open source packages to be deterministic (fully reproducible).

Unfortunately, you have no enough resource to achieve this at current time. So, please keep this issue open in the hope that you will get enough resources in future to achieve it .....

Best.

saghul:
Sorry, we can't sspend more resources on making yet anothher package.

Nokia808:
@saghul 
Never mind. It is not a big problem as long as built-in update feature of your AppImage is working regularly. This way, user need for only one time verification of heir/his download & the remaining will be automatic .....
I had very bad experience with built-in automatic update of Rambox. But, I feel your AppImage is perfectly implement this feature .....

digitalethics:
@Nokia808 Since when is there an update feature in AppImages? Can you point me to this in their repo?

digitalethics:
> Sorry, we can't sspend more resources on making yet anothher package.

Would you mind re-opening this issue and add a "need help" label so someone can pick this up? I think this is also a matter of outreach and it should be fairly trivial to package this for someone experienced.

PR #446, opened 2020-09-26
Flatpak from Flathub link

ovari:


saghul:
I don't know who publishes that so I'm not very comfortable recommending users to install it.

ovari:
@ananace can you please make yourself known to @saghul so that https://github.com/flathub/org.jitsi.jitsi-meet/ can be added to [README.md](https://github.com/jitsi/jitsi-meet-electron/blob/master/README.md)?

ananace:
Personally, I'd not call it a "Flathub download" as Flathub is just a repository and not the download, just the word "Flathub", "Flatpak", or something like "Available on Flathub" would be a much better way to describe it.
Perhaps marking it as third-party would be for the best too, to make it even more clear that any Flatpak-specific issues should be directed to there. (Unless the Jitsi team wishes to also take maintainership of the Flatpak, which they're more than welcome to)

On a side note, the Flatpak is done to require - and cause - as little work for here as possible; it's basically just taking the regular AppImage, unpacking it into the sandbox, fixing the desktop exec and icon, and adding some metainfo/appdata XML as required by Flathub.

ovari:
> Personally, I'd not call it a "Flathub download" as Flathub is just a repository and not the download, just the word "Flathub", "Flatpak", or something like "Available on Flathub" would be a much better way to describe it.

Thanks, fixed.

> I'm not very comfortable recommending users to install it.

@saghul would @ananace giving you full access to https://github.com/flathub/org.jitsi.jitsi-meet/ help?

ananace:
And as a note on that; I'm not the person to give out that access, that would have to be from the Flathub people - who are more than happy to transfer ownership to the upstream teams that wish to handle their own packages.

ovari:
Can you please add the Flathub people to this issue? Thank you

ananace:
I'd quite like to wait for the Jitsi team to decide that they wish to take ownership of the Flathub packaging before dragging them into that.

The reason I did the Flathub addition in the first place was because there had been requests (#320, #426) for a Flatpak package before, which were closed due to lack of time to spend on it. I'd not wish to force them to add that extra bit of burden if that's still true.

ovari:
https://github.com/flathub/org.jitsi.jitsi-meet/issues/4#issuecomment-699558884

@saghul can you please accept the patch?

saghul:
What patch? Just to be clear, while I'm happy to take a patch that would make this easier for you, we have no resources to devote to this, so while we can add it to the README it will be with a "Third-Party builds" heading or something similar.

ovari:
Sorry, there was a patch but then we added the flathub/flatpak attempt to translate the Desktop file which GitHub advised didn't work. We are doing this all in the web-based interface. It didn't work so we deleted it.

PR #447, opened 2020-09-27
Update package.json

ovari:
App description in Linux


ovari:
Sample of `App description in Linux` using `desktop` file 
https://github.com/pdfarranger/pdfarranger/blob/master/data/com.github.jeromerobert.pdfarranger.desktop

@saghul @ananace would a similar file at 
https://github.com/jitsi/jitsi-meet-electron/blob/master/data/org.jitsi.jitsi-meet.desktop 
work, instead of updating the `package.json` file?

Thank you

PR #490, opened 2020-12-02
fix: Disable hardware acceleration on Windows, it causes screenshare to flicker.

jallamsetty1:


main.js
```diff
@@ -23,12 +23,18 @@ const URL = require('url');
...
 const pkgJson = require('./package.json');
+const OS = require('os');
 
 const showDevTools = Boolean(process.env.SHOW_DEV_TOOLS) || (process.argv.indexOf('--show-dev-tools') > -1);
 
 // We need this because of https://github.com/electron/electron/issues/18214
 app.commandLine.appendSwitch('disable-site-isolation-trials');
 
+// We need to disable hardware acceleration on Windows because its causes the screenshare to flicker.
+if (OS.platform() === 'win32') {
```
saghul, line 34: Can you use `process.platform` instead, since it's already imported? Note how we use it to not close the last window on macOS.

saghul (changes requested):

main.js
jallamsetty1: oh I didn't see that part, thanks for pointing it out.

saghul (approved):

Issue #477, opened 2020-11-16
The "Give feedback" button within the Help menu does nothing


netsrotr:
## Current behavior
The "Give feedback" button within the Help menu does nothing

## Expected Behavior
It should point me to the issues GitHub page, or any other Jitsi App forum site.

# Environment details
Jitsi Electron App v2.4.2

saghul:
Fixed in https://github.com/jitsi/jitsi-meet-electron/commit/01c2468666536496164bb4feb47161fe530a4bca

PR #498, opened 2020-12-26
Update openExternalLink.js

semisenex:
Allows opening mailto links.
Fix issue [#477](https://github.com/jitsi/jitsi-meet-electron/issues/477#issue-743628544)


saghul:
Nice catch! Actually, I'd rather have a link pointing to the issues tab of this repo. Mind changing that instead?

saghul:
Replaced in 01c2468666536496164bb4feb47161fe530a4bca

PR #500, opened 2021-01-03
Update Electron to fix i18n on macOS

csett86:
Since Electron 10.1.6 i18n is broken on macOS and app.getLocale()
always return en_US, see https://github.com/electron/electron/issues/26612

Updating Electron to 11.1.1 fixes the issue and app.getLocale() again
returns the proper values.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7512,9 +7512,9 @@
       }
     },
     "electron": {
-      "version": "10.1.6",
-      "resolved": "https://registry.npmjs.org/electron/-/electron-10.1.6.tgz",
-      "integrity": "sha512-Wyiq5Fy64KAa51i72m+5zayYKSm9O5lnittUdaElAn3PAzGl3yDifYO2QsXR7k/iKxWVSROOPzf43mXYytL67Q==",
+      "version": "11.1.1",
+      "resolved": "https://registry.npmjs.org/electron/-/electron-11.1.1.tgz",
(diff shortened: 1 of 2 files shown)


saghul:
Last I tried updating to Electron 11 there was an error in our logger (since fixed) which prevented Jitsi Meet from working. Have you experienced this? Note that meet.jit.si is always ahead of stable, so even if it works there we need to wait for a new stable release not to break people's installations.

csett86:
Hi @saghul, I have tested this successfully with both current meet.jit.si and the current stable release of jitsi-meet that I have running on my on install, both working fine.

saghul:
Oh, maybe they fixed it on the Electron side, nice,

nodeg:
Hi @csett86, @saghul I did some quick testing with the most recent version of Electron (11.2.1) on macOS Big Sur. Everything worked fine on several Jitsi servers. The change in the `package.json` is so small that I thought I leave some comment here instead of opening another PR/issue since this PR is not merged already.

saghul:
Thanks for the comment. I've been quite busy these past weeks and haven't been able to come back to this yet.

z1r0-:
FYI: Electron 11 also supports Apple Silicon. I tried to build it with electron 11.2.1 and target arch "arm64". It failed building the dmg file (maybe I'm missing some dependencies or something), but app-file was build successfully and seems to run without issues.

image

saghul:
Updated in 2e482a132cf392303fc89281c76dc54bb40ac7dc

PR #505, opened 2021-01-14
Add enable remote control to settings panel

ojehle:
enable remote control is disabled by default. 

if you run your own instance of jitsi, then the remote control can be enabled in my opinion.  

Perhaps it could be protected a little bit more by checking, that the default server is overriden.


saghul:
Sorry, we cannot do this. Check the security advisory.

PR #521, opened 2021-02-23
Added Dutch translation

cedricroijakkers:
I've added a Dutch translation file, based on the wording in the jitsi-meet app.


saghul (approved):
Dank je wel!

PR #528, opened 2021-03-03
Electron 12

saghul:


main.js
```diff
@@ -29,9 +29,6 @@ const showDevTools = Boolean(process.env.SHOW_DEV_TOOLS) || (process.argv.indexO
 // We need this because of https://github.com/electron/electron/issues/18214
 app.commandLine.appendSwitch('disable-site-isolation-trials');
 
-// We need to disable hardware acceleration because its causes the screenshare to flicker.
-app.commandLine.appendSwitch('disable-gpu');
```
saghul, line 33: @jallamsetty1 We should be good here, because Electron 12 is M89 right?

jallamsetty1: Yes.

saghul: Cool because I blindly went ahead :-P

PR #534, opened 2021-03-11
Enhancements to make it Mac App Store (mas) compatible

csett86:
- disable the autoupdater if running as mas (was not working anyway, just logging an error on every start)
- replace check via app.requestSingleInstanceLock() with LSMultipleInstancesProhibited in Info.plist
  due to https://github.com/electron/electron/issues/15958
- Ignore certificate-error as otherwise every network connection (eg. Jitsi URL in iframe) is cancelled
- Quit the app also when all windows are closed to conform to macOS Human Interface Guidelines
  Comments from review:
  If the application is a single-window app, it might be appropriate to save data and quit the app when the main window is closed.
- "asarUnpack": "**/*.node" to also sign the native addons when packaging
- add the required mas-specific entitlements which include the app-sandbox key

### DIFF
--- a/main.js
+++ b/main.js
@@ -161,7 +161,9 @@ function createJitsiMeetWindow() {
     setApplicationMenu();
 
     // Check for Updates.
-    autoUpdater.checkForUpdatesAndNotify();
+    if (!process.mas) {
+        autoUpdater.checkForUpdatesAndNotify();
+    }
 
     // Load the previous window state with fallback to defaults.
     const windowState = windowStateKeeper({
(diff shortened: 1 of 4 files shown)


csett86:
FYI: These changes are the upsteaming of https://apps.apple.com/de/app/digitales-klassenzimmer/id1508736201

saghul:
Thanks for this! We have no intention of publishing the app in the store but hopefully this helps others who want to!

I'm concerned about the certificate error ignoring. There must be a way to properly validate certificates, otherwise it's a pretty big security flaw.

csett86:
I agree and I'll try to get the certificate validation working as I also see this as an issue, but for now I don't have a better workaround.

csett86:
Unfortunately the certificate errors are not specific to this app, I reproduced them with the minimal https://github.com/szwacz/electron-boilerplate sample app by including a css in the app/app.html and then sandboxing the app.

DNS resolution works fine, but the embedded chromium cannot verify any TLS certificates and all requests from the browser window fail with `Failed to load resource: net::ERR_CERT_INVALID`.

csett86:
Followup is https://github.com/jitsi/jitsi-meet-electron/pull/537

csett86:
Opened an upstream issue with a minimal repro for the certificate issue with https://github.com/electron/electron/issues/28162

saghul:
Thank you!

Issue #508, opened 2021-01-22
Screen sharing previews broken


ubmarco:
## Description
---

The screen share preview window shows broken images.

## Current behavior
---

![image](https://user-images.githubusercontent.com/839934/105465757-a13cdb00-5c93-11eb-99c7-e7e6d72d61e1.png)

![image](https://user-images.githubusercontent.com/839934/105465892-ce898900-5c93-11eb-8628-dd11a93e18f8.png)

## Expected Behavior
---
The screen share preview window shows the window's contents.

## Possible Solution
---

## Steps to reproduce
---
Click on the screen share button.
# Environment details
---
Manjaro Linux 20.2.1, 64bit, jitsi-meet-electron version 2.4.2

saghul:
Cannot reproduce this.

ubmarco:
Hmm ok. Maybe I can provide some logs? When looking into the debug toolbar, I see a lot of requests of the following type when opening the screen share window:
![image](https://user-images.githubusercontent.com/839934/108842890-3ce6a180-75da-11eb-83e0-e468ddd189fe.png)
Here is the response tab:
![image](https://user-images.githubusercontent.com/839934/108842928-4a9c2700-75da-11eb-8387-a6c6a4f5d149.png)
These are base64 encoded png images.

saghul:
Those are provided by Electron itself, we just render them. Any chance you are using wayland?

ubmarco:
No, I'm using X11.
```
marco@tux01 ~ loginctl
SESSION  UID USER  SEAT  TTY
      2 1002 marco seat0    

1 sessions listed.

marco@tux01 ~ loginctl show-session 2 -p Type
Type=x11
```
This is my `neofetch` output:
![image](https://user-images.githubusercontent.com/839934/108863262-4d564680-75f1-11eb-8a7e-be99ddddbe4b.png)

saghul:
That's odd. Can you try the freshly released 2.5 version?

ubmarco:
I updated to 2.5.1 using the package https://aur.archlinux.org/packages/jitsi-meet-desktop-bin/. Still the same issue.
Are there any logs I can investigate?
The screen share preview works perfectly in Chromium.

saghul:
Please use our release not AUR to test it.

ubmarco:
[...]

PR #536, opened 2021-03-12
Document #508

t2d:


saghul:
This blocks unsafe eval, which blocks loading WASM, doesn't it? That prevents E2EE from working.

t2d:
No, E2EE seems to work just fine.

saghul:
Excellent, cheers!

PR #540, opened 2021-03-13
Switch mac build to universal build

csett86:
Apple's documentation suggests that apps should be shipped as universal binaries
to simplify the process for the users. See eg. https://developer.apple.com/documentation/apple-silicon/porting-your-macos-apps-to-apple-silicon

Switch back to that simpler world and adapt the README accordingly.

Also update electron-updater to latest version (matching electron-builder version)
as part of this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,9 +17,9 @@ Desktop application for [Jitsi Meet] built with [Electron].
 
 Download our latest release and you're off to the races!
 
-| Windows | macOS | macOS (Apple Silicon)| GNU/Linux (64-bit only) |
-| -- | -- | -- | -- |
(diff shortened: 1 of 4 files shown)


saghul:
Oh, brilliant! I tried universal builds first, but didn't figure out the asar file trick! I'll give it a quick test and land. Thank you!

csett86:
You're welcome, I have this asar rewrite (albeit without the runtime detection) running fine on github.com/csett86/jitsi-meet-electron and github.com/freifunkMUC/jitsi-meet-electron for some weeks now and no complaints so far.

Looking forward to get this upstreamed, with the additional runtime detection so that it does not break if someone changes the arch in package.json down the road (eg. to arm64 only in some years).

saghul:
What do you mean by runtime detection?

csett86:
Previously I just hardcoded the rewrite based if the app is running on mac OS (and my that assuming that arch in package.json is set to universal):

https://github.com/csett86/jitsi-meet-electron/blob/main/main.js#L179-L185

With this pull-request I at least check at runtime if the arch-specific `app-arm64.asar` file exist before rewriting the basePath:

https://github.com/csett86/jitsi-meet-electron/blob/universal/main.js#L179

saghul:
Gotcha, makes sense!

saghul (approved):

PR #546, opened 2021-03-23
Update openExternalLink.js

bsimml:
Change RegEx so you can open the 'mailto' link used by the button 'Your Default Email'

### DIFF
--- a/app/features/utils/openExternalLink.js
+++ b/app/features/utils/openExternalLink.js
@@ -2,7 +2,7 @@ const { shell } = require('electron');
 const url = require('url');
 
 
-const protocolRegex = /^https?:/i;
+const protocolRegex = /^(https|mailto)?:/i;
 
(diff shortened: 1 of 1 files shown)


app/features/utils/openExternalLink.js
```diff
@@ -2,7 +2,7 @@ const { shell } = require('electron');
 const url = require('url');
 
 
-const protocolRegex = /^https?:/i;
+const protocolRegex = /^(https|mailto)?:/i;
```
saghul, line 5:
```suggestion
const protocolRegex = /^(https?|mailto):/i;
```

Untested, but isn't this what we want?

bsimml: Yeah your right this is what we want. Made a typo.

saghul (approved):

Issue #231, opened 2020-04-07
Linux: 2.0.0 fails to start


monga:
[193209:0407/164939.241277:FATAL:setuid_sandbox_host.cc(157)] The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /tmp/.mount_jitsi-KCGvvK/chrome-sandbox is owned by root and has mode 4755.

I was a happy user of the previous version, but now it fails with the above error.
I'm on an Debian unstable system (x86_64),  with kernel 5.5.1

As a workaround, I found out (see https://github.com/electron/electron/issues/17972)  that 

```sh
sudo sysctl kernel.unprivileged_userns_clone=1
```
makes it work again.

saghul:
Thanks for the report!

btlogy:
I've seen the bug while building from master yesterday.
But I was too busy getting v1.1.1 running on many devices to start digging...
Thanks to the auto-update feature, all those devices are now failing on this FATAL error after updating in the background :-/

```
FATAL:setuid_sandbox_host.cc(157)]` The SUID sandbox helper binary was found, but is not configured correctly. Rather than run without sandboxing I'm aborting now. You need to make sure that /tmp/.mount_jitsi-tMA386/chrome-sandbox is owned by root and has mode 4755.
Trace/breakpoint trap
```

I'm using mostly Debian 10 stable with kernel 4.19.98-1.

Thanks @monga for the workaround. :+1: 
It helps me too.

And thanks everyone else to be working on improving this app. :+1: 
Got most of my family using it since the lock down.
And the electron app is much easier and perform better than the web version.

REM: I'm getting something working on Raspbian too, but with many issues to fix (unstable cam, weird noisy sound and abi error while packaging).

gerroon:
The thing about the work around is that you are going against the Debian's security settings.  And one still needs the admin rights to get that fix.

I do hope that there is a fix with this during appimage packaging.

saghul:
I could reproduce it with Ubuntu 19.10 by setting `sudo sysctl kernel.unprivileged_userns_clone=0` myself, since it seems to default to 1.

Hopefully we can solve this quickly, but if anyone has ideas on how to fix it, they'd be more than welcome.

saghul:
Tried to fix it on https://github.com/jitsi/jitsi-meet-electron/pull/235 and made a beta 4, but alas the problem persists :-/

[...]

PR #550, opened 2021-03-28
Fix running AppImage on Debian 10+ (#231 )

csett86:
Move the real binary to jitsi-meet.bin and create a wrapper script (thanks to https://github.com/electron-userland/electron-builder/issues/5371#issuecomment-791771150) that contains the --no-sandbox and is also used in the AppImage.

With this we can remove the documented workaround and the AppImage runs fine on Debian 10+ as well.

The patch for app-builder-lib can also be dropped, as the desktop file now calls the wrapper script which then adds the --no-sandbox as a param when executing the actual binary.


csett86:
@saghul This pull request is now ready for review. I have squashed the commits in the end as it was a bit of back and forth that history is not really relevant to be kept for later.

Looking forward to get this reviewed as this finally removes the manual workaround of launching with --no-sandbox on Debian 10+.

saghul (approved):
LGTM, thanks for fixing this! I'll give it a quick test.

csett86:
@saghul This breaks the jitsi-meet:// protocol handling, I will rework it. Please dont merge yet!

csett86:
@saghul jitsi-meet:// protocol handling now also continues to work (as it should), the order of args in the wrapper script was simply wrong...

saghul (approved):

Issue #543, opened 2021-03-15
Icon isn't correctly packaged


daenney:
## Description
A recent (I think the 2.6) update to Jitsi has caused it to lose its icon in the application launcher.

## Current behavior
Generic application icon is shown when using the application launcher

## Expected Behavior
Jitsi icon is shown

## Possible Solution
Correctly package the app icon. When unpacking the AppImage, the icon is in the wrong spot:

```
squashfs-root/usr/share/icons/hicolor
squashfs-root/usr/share/icons/hicolor/0x0
squashfs-root/usr/share/icons/hicolor/0x0/apps
squashfs-root/usr/share/icons/hicolor/0x0/apps/jitsi-meet.png
```

0x0 icons aren't valid/recognised by desktop environments. However, the icon itself is 512x512, and should be placed in the appropriate directory:

```
file /usr/share/icons/hicolor/0x0/apps/jitsi-meet.png 
/usr/share/icons/hicolor/0x0/apps/jitsi-meet.png: PNG image data, 512 x 512, 8-bit/color RGBA, non-interlaced
```

This is usually an electron-builder/packaging issue: https://www.electron.build/icons#linux and might require upgrading electron-builder.

# Environment details
---
* Linux: 5.11.6
* GNOME: 3.38.4

csett86:
I have provided a fix in pr #551 for this issue.

saghul:
Fixed in https://github.com/jitsi/jitsi-meet-electron/pull/551

PR #551, opened 2021-03-28
Linux: Fix Desktop Icon (#543)

csett86:
This is a workaround for the electron-builder issue
https://github.com/electron-userland/electron-builder/issues/5294


saghul (approved):

saghul:
Thank you!

PR #555, opened 2021-04-04
chore(deps) Update dependencies

csett86:
devDependencies left out intentionally, as updating babel & eslint requires
more work and more dependencies due to the deprecation of babel-eslint.


saghul (approved):
Thank you for doing this!

PR #561, opened 2021-04-17
Adding sq translation

ujdhesa:
Here you are a translation for sq locale. 

TIA


saghul:
You need to add it here too: https://github.com/jitsi/jitsi-meet-electron/blob/master/app/i18n/index.js

saghul:
There is a syntax error check the trailing commas

ujdhesa:
> There is a syntax error check the trailing commas
I hope that was the right to blame.

saghul:
Now you're missing the comma after the nl locale. Please test your code.

ujdhesa:
> Now you're missing the comma after the nl locale. Please test your code.
I apologise for having overlooked again.

saghul (approved):

PR #574, opened 2021-05-05
deps: Update electron-builder to 22.11.1

csett86:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,9 +5,9 @@
   "requires": true,
   "dependencies": {
     "7zip-bin": {
-      "version": "5.0.3",
-      "resolved": "https://registry.npmjs.org/7zip-bin/-/7zip-bin-5.0.3.tgz",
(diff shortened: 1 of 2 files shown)


saghul:
Any interesting features we're getting out of this?

csett86:
Mostly small bugfixes like "DMG assembly on Apple Silicon", "extend http file limit" and the possibility to specify the package name for linux packages (like deb), but nothing really important.

So I consider it a small update just to keep up to date what everyone else is using.

saghul (approved):

saghul:
Hum, I just tried to make a new release and the artifacts are not being published to the draft release. Have you observed this? This dependency is the only relevant one I think.

csett86:
@saghul yes, I can reproduce the issue as well. That seems to be a regression from that electron-builder release. Please excuse the hassle, I suggest to revert this pull request for now.

saghul:
👍

Issue #564, opened 2021-04-20
Disable automatic gain control


alxpettit:
Jitsi Meet Electron app, when open, will often cause my microphone audio to fly up and down when I don't want it to. This is due to Electron's automatic gain control. Please add an option to disable this.

tio-trom:
Excuse me but where is this setting in Jisi? I honestly cannot find it...

saghul:
It's in the electron app settings panel.

tio-trom:
Ahh yes...I was looking at the Jitsi Meet settings after I created a call. Thanks!

PR #582, opened 2021-05-11
feat(agc): Add setting to disable automatic gain control

csett86:
In some OS/Chromium combinations the automatic gain control goes slightly
crazy, but normally its fine. Thus keep the default as is, but add an option
for the users to disable it if required.

Closes: #564

### DIFF
--- a/app/features/conference/components/Conference.js
+++ b/app/features/conference/components/Conference.js
@@ -35,6 +35,11 @@ type Props = {
     _alwaysOnTopWindowEnabled: boolean;
 
     /**
+     * Disable automatic gain control.
+     */
+     _disableAGC: boolean;
+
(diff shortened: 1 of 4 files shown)


app/features/conference/components/Conference.js
```diff
@@ -63,6 +63,11 @@ type Props = {
      * Start with Video Muted.
      */
     _startWithVideoMuted: boolean;
+
+    /**
+     * disable automatic gain control
+     */
+    _disableAGC: boolean;
```
saghul, line 70: CAn you sort this prop alphabetically?

csett86: Sure, good remark, done

PR #583, opened 2021-05-12
ci: move to node 14 (LTS)

csett86:
Next electron builder will require node 14 for eg "fs/promises"


csett86:
@saghul FYI: I would merge these PRs myself (or directly push to master) after successful testing to help you with maintenance, but unfortunately thats rejected for contributors with the current permission setup for this repo.

csett86:
@saghul I have tested this including updating and all works fine as expected.

csett86:
Once this is in, I have a followup PR for an electron patch update (just chromium patch updates)

Issue #567, opened 2021-04-28
Use electron with Pipewire to support screensharing under Wayland


snaggen:
## Description
After Electron 8.2.2 version the rtc_use_pipewire=true build flag to support Wayland screen sharing
So by making sure the builds you provide is based on electron with this build flag, you should get proper screen sharing support under Wayland. And this is probably going to be more of an issue now, when Ubuntu released with wayland as default.

snaggen:
Ok, before electron v12 it was only in an experimental branch, but electron v12 should contain the pipewire fixes with RTC_USE_PIPEWIRE build flag according to the all mighty Internet.

saghul:
My understanding is there are 2 command line flags you can use when launching the app to get Wayland support. I'm not 100% sure this fixes screen-sharing.

We are not going to be making custom Electron builds, however. Once upstream Electron has it, we'll use it.

snaggen:
It seems that official builds of electron now should have the rtc_use_pipewire enabled by default according to https://github.com/electron/electron/issues/23063#issuecomment-744339286
In that thread it is claimed that the flag --enable-features=WebRTCPipeWireCapturer is required, so I tested this with the appimage. Which then just give me a broken image instead of a window
![bild](https://user-images.githubusercontent.com/6420639/116455738-ba8c8f00-a861-11eb-9392-04a081345644.png)

saghul:
Did you also set these? 

--enable-features=UseOzonePlatform --ozone-platform=wayland

snaggen:
No, I originally only used --enable-features=WebRTCPipeWireCapturer , since I know that in Chromium the use of Pipewire is not dependent on Wayland. I have working WebRtc in Chromium through pipewire using XWayland in Fedora 33/34. So I assume this is far more stable then the rest of the wayland related changes. 
But I tested to also add the flags you suggested --enable-features=UseOzonePlatform --ozone-platform=wayland , and that gave me a Wayland version (but since they do not support CSD yet, I got no decorations) but now the screensharing was back to X only strangely enough. In the application windows view, I see only my X applications listed. That was not what I expected.

saghul:
It's labeled as experimental AFAIK, so it may take a few more versions to get there I'm afraid.

snaggen:
[...]

PR #584, opened 2021-05-12
feat(linux): enable PipeWire support for screensharing

csett86:
This is required on wayland based desktops like Fedora 34 or
Ubuntu 21.04.

Closes: #567

### DIFF
--- a/main.js
+++ b/main.js
@@ -37,6 +37,9 @@ app.commandLine.appendSwitch('disable-features', 'IOSurfaceCapturer');
 // Enable Opus RED field trial.
 app.commandLine.appendSwitch('force-fieldtrials', 'WebRTC-Audio-Red-For-Opus/Enabled/');
 
+// Enable optional PipeWire support.
+app.commandLine.appendSwitch('enable-features', 'WebRTCPipeWireCapturer');
+
 // Needed until robot.js is fixed: https://github.com/octalmage/robotjs/issues/580
 app.allowRendererProcessReuse = false;
 


csett86:
Tested this successfully from the AppImage on Fedora 34 Workstation.
Also no regression on X11 based desktops, verified with Xubuntu 20.04, and neither regressions on other platforms (mac, win) due to the flag.

vchernin (approved):

vchernin:
While this implementation does have sometimes have bugs, it is a definite improvement over the nonexistent screen sharing support in wayland from before. Also, [others have been working on fixing upstream issues](https://jgrulich.cz/2020/12/18/webrtc-chromium-updates-in-2020/).

Sidenote, I [could not find any regressions](https://github.com/flathub/com.slack.Slack/pull/118#issuecomment-833970195) when [making an equivalent change on the Slack Flatpak](https://github.com/flathub/com.slack.Slack/pull/118/commits/8552181dc907f00b9682c784c8876205098d3617).

saghul (approved):

yu-re-ka:
This broke my usage of element-desktop: I used to pass `--enable-features=UseOzonePlatform --ozone-platform=wayland` to the electron process to enable the wayland backend.

vchernin:
That issue is probably the same as https://github.com/flathub/org.chromium.Chromium/pull/131#issuecomment-884612212. `--enable-features` isn't able to dealt with multiple arguments well.

I think it's reasonable enough to change the argument as in #630 (assuming that works as intended). @yu-re-ka would you mind if I reused #630 in Element?

Oh and for the record I do think ` --enable-features=UseOzonePlatform` won't be necessary in the future, it will be enabled by default anyhow. But it seems reasonable to not interfere if users pass command line options.

yu-re-ka:
> @yu-re-ka would you mind if I reused #630 in Element?

Of course not, see https://github.com/vector-im/element-desktop/pull/261 😄

vchernin:
> Of course not, see https://github.com/vector-im/element-desktop/pull/261 😄 

Oh, thanks for fixing that too! 👍

PR #586, opened 2021-05-22
fix(mac): remove wifi stats as they cause hangs every 10 seconds

csett86:
The getWiFiStats causes screen freezes every 10 seconds for 2-4 seconds,
thus temporarily remove them.

Closes: #595

### DIFF
--- a/app/features/conference/components/Conference.js
+++ b/app/features/conference/components/Conference.js
@@ -266,7 +266,11 @@ class Conference extends Component<Props, State> {
             setupAlwaysOnTopRender(this._api);
         }
 
-        setupWiFiStats(iframe);
+        // Disable WiFiStats on mac due to jitsi-meet-electron#585
+        if (window.jitsiNodeAPI.platform !== 'darwin') {
+            setupWiFiStats(iframe);
+        }
+
         setupPowerMonitorRender(this._api);
     }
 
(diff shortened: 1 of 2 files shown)


csett86:
freeze-wifi-stats

damencho:
Have you found that reason for the hang? That just executes `netsh.exe wlan show interfaces` ...

csett86:
Unfortunately I have not found the reason for the hang, but it is reproducable with current macOS 11.3.1 with both current jitsi-meet-electron 2.8.6 and eg. 2.6.0. Older versions like 2.4.2 do not show the behaviour.

It may be a mac-only issue, but from looking at the code in jitsi-meet-electron-utils I could not determine what could cause the hangs, especially as the code hasn't changed since long and also just executes 2 cli calls which execute quickly when copied locally.

csett86:
I have further tested and the issue only appears on mac, windows and linux are not affected. Thus I updated the pull request to only target mac.

Unfortunately the root cause for the hang is still unknown to me.

damencho:
@csett86 are you still able to reproduce? If you repro do you see some errors in system.log?
Errors like `(com.apple.xpc.launchd.domain.pid.networksetup.NNNNN): Failed to bootstrap path: path = /usr/sbin/networksetup, error = 2: No such file or directory}}`?
Thank you.

damencho (approved):

PR #589, opened 2021-05-24
refactor: reduce shipped and installed size

csett86:
This moves all webpacked deps to be automatically included from the asar
and thus reduces the shipped binary by 14MB on Windows, 16 MB on Mac and
12MB on Linux. The installed size on mac is reduced by 105 MB.

Previously all the dependencies were webpacked and minified in the asar in the
build folder and in its full installed version unused under node_modules.

Only keep the dependencies that are externally required in the node-modules
folder.


csett86:
@saghul FYI: I would merge these PRs myself (or directly push to master) after successful testing to help you with maintenance, but unfortunately thats rejected for contributors with the current permission setup for this repo.

csett86:
@saghul I have tested this successfully on windows, mac and linux. Please approve if you agree to the approach

saghul (approved):

PR #594, opened 2021-05-31
feat: support config overwrite via URL parameters

csett86:

### DIFF
--- a/app/features/conference/components/Conference.js
+++ b/app/features/conference/components/Conference.js
@@ -209,6 +209,15 @@ class Conference extends Component<Props, State> {
         const roomName = url.pathname.split('/').pop();
         const host = this._conference.serverURL.replace(/https?:\/\//, '');
         const searchParameters = Object.fromEntries(url.searchParams);
+        const hashParameters = url.hash.substring(1).split('&')
+            .reduce((res, item) => {
+                const parts = item.split('=');
+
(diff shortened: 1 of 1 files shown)


csett86:
I tested this successfully with the examples from #67 like config.disableAP=true&config.disableAEC=true&config.disableNS=true&config.disableAGC=true&config.disableHPF=true&config.stereo=true

I chose to put it after the app settings so that one can individually override for a conference the settings done in general in the app.

csett86:
Currently not supported is the nested ones like config.p2p.*

saghul (approved):

rogeriomq:
In the jitsi-meet project, there is a function that does this parse and returns a complete object that comes via hashParams([src/react/features/base/util/parseURLParams.js](https://github.com/jitsi/jitsi-meet/blob/da7358d5648175c546878e6ec3889d68e93ed8b9/react/features/base/util/parseURLParams.js)).
I made use of parseURLParams, and besides props for configOverwrite, I also retrieve other custom keys and also userInfo.

I will try to replicate and perform a pullRequest.
But the idea is already there.  😊

bilogic:
Is there a way to tell if AP has been disabled in the electron app? Or must we rely on our ears?

saghul:
You could open the dev console (same shortcut as in a Chrome browser) and check.

PR #605, opened 2021-06-29
chore(deps): update electron to 12.0.13

csett86:
Security fixes for bundled chromium

For all details see https://github.com/electron/electron/releases/tag/v12.0.13

Signed-off-by: Christoph Settgast <csett86@web.de>

PR #593, opened 2021-05-27
refactor: remove dependency to deprecated remote module

csett86:
the remote module will be removed from electron 14 onwards,
so replace the locale detection with native browser api that
is available in the renderer.

Signed-off-by: Christoph Settgast <csett86@web.de>

PR #606, opened 2021-06-30
2.8.8

csett86:


csett86:
@saghul I would like to cut a release to including the following:

9b1e390 refactor: reduce shipped artifact size (#589)
e5d5bcc feat: support config overwrite via URL parameters (#594)
defb570 feat(recent-list): add an explanatory label (#591)
a028aef lang(en): clarify that main input is room name, not participant (#592)
4171437 chore(deps): update electron to 12.0.13 (#605)
0fdb225 fix(deb): match dependencies with ldd from electron 12 (#604)
2baa4b5 refactor: remove dependency to deprected remote module (#593)

saghul (approved):

PR #621, opened 2021-08-10
chore(deps): update electron to 13.1.9 

csett86:
@saghul Tested audio, video and screensharing, all working as expected.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8064,9 +8064,9 @@
       }
     },
     "electron": {
-      "version": "13.1.7",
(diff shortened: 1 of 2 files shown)


saghul:
Can we please hold the Node 14 update?

We haven't updated our environments for Jitsi Meet and this would make it kinda annoying.

csett86:
Understood, I was not aware of that dependency. I have removed the electron-builder and nodejs update from this PR. Retested and everything still works fine (as expected).

saghul (approved):
Thank you!

PR #633, opened 2021-09-12
chore(deps) Update external_api.js

csett86:
Includes

fix(misc) fix typeof comparison to undefined
feat(errors) Expose errors through Iframe API (#9801)
fix(external-api): Avoid naming event 'error'
feat(external_api): Command to set participant volume


saghul:
None of those changes are used by the Electron wrapper really.

csett86:
Yes, the contained fixes were an attempt to fix issues with an older jitsi-meet install and password-protected rooms, but its anyway not related to that, thus closing this.

PR #635, opened 2021-09-16
BREAKING CHANGE: Switch to local storage

csett86:
electron-storage with its current version is unsupported on Electron 14+ (due to
remote module removal), electron-storage 7+ is really slow (10 seconds on start),
thus switch to local storage.

Unfortunately all settings are lost, as migrating from electron-storage
to local storage is not feasible without continuing to have slow initial
startup times.


saghul:
Is there no electron-storage version that works? Wow.

saghul (approved):

saghul (approved):

saghul:
Let's bump the major then, since people will lose their display name.

csett86:
I found a way to migrate the settings (and thus avoid the breakage), then we can at least ship a version with electron 13 and migrate to local storage and only later on remove electron-store when updating to electron 14.

I suggest a 3 months migration period before removing electron-store and the migration.

See #638 for all the details, closing this breaking change in favor of it.

saghul:
Excellent!

PR #640, opened 2021-09-22
Update package.json

### LINKED ISSUE #636: Desktop file description/comment localization: hu (Hungarian - magyar)
How can `Jitsi Meet Desktop App` in the Desktop file be translated?

The Hungarian translation is `Asztali Jitsi Meet-alkalmazás`

The image below is with the Flathub version running on the Linux Mint 20.2 Cinnamon operating system.

Thank you

kép

ovari:
Closes #636, closes https://github.com/electron-userland/electron-builder/issues/6266

Seems that `description` should be `Comment` as per https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html

Hope this works.

Thank you

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,8 +58,9 @@
     },
     "linux": {
       "artifactName": "jitsi-meet-${arch}.${ext}",
-      "category": "VideoConference;AudioVideo;Audio;Video;Network",
-      "description": "Jitsi Meet Desktop App",
+      "Comment": "Jitsi Meet Desktop App",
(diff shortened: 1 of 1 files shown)


ovari:
@mmaietta the key-value pairs are not [simple](https://github.com/electron-userland/electron-builder/issues/6266#issuecomment-924571000) for me. Can you please advise how to fix this merge request? Thank you

csett86:
@ovari If your editor (or the online github editor) does not support json validation, I suggest to use something like https://jsonlint.com/ so that its at least valid json.

PR #651, opened 2021-11-01
ci: move to macOS 11 and 20.04

csett86:


csett86:
@saghul: Tested the results successfully on mac and linux (as windows is unchanged). Please review and merge.

saghul (approved):

PR #655, opened 2021-11-03
Update jitsi-meet-electron-utils to be ready for electron 14+

csett86:


csett86:
Tested successfully, I will change to a commit on master of jitsi-meet-electron-utils once https://github.com/jitsi/jitsi-meet-electron-utils/pull/134 is merged.

saghul (approved):

saghul (changes requested):
Let's guard it with a constant check, we don't want this to eventually be enabled by mistake.

csett86:
Agree, will add a guard similar to the Conference.js guard for remotecontrol

csett86:
Now properly guarded for main process as well

saghul (approved):

csett86:
I will dig deeper why the windows builds are now both failing with npm8 / node 16, both here and https://github.com/jitsi/jitsi-meet-electron-utils/pull/135 with the same error.

Until then not going to merge this, as otherwise we no longer have a windows build...

csett86:
The windows build is now fixed here as well.

saghul (approved):

PR #658, opened 2021-11-05
fix(mac): electron-updater broken on mac

csett86:
Since release 2.9.1 autoupdate was broken (only!) on mac, due to
https://github.com/electron-userland/electron-builder/pull/6149 being
part of the used version of electron-updater

Revert back to stable electron-updater.

Also directly prepare hotfix release 2021.11.2

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "jitsi-meet-electron",
-  "version": "2021.11.1",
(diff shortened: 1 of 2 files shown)


csett86:
@saghul I successfully tested this on mac as well.

saghul (approved):
The Linux failure looks like a network glitch so go ahead.

PR #660, opened 2021-11-07
feat: as robotjs is context-aware, allow reuse of render processes

csett86:


saghul (approved):

saghul:
Code LGTM, but something is wrong on Windows?

csett86:
Rebuild without change worked, so maybe a fluke

PR #666, opened 2021-11-13
ci: simplify thanks to prebuilds

csett86:
as @jitsi/robotjs and @jitsi/electron-sdk now have prebuilds, no need for build
tools here

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,8 +6,6 @@ on:
   push:
     branches:
       - master
-    tags:
-      - '*'
   pull_request:
     branches:
       - master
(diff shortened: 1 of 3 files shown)


saghul:
Looks like Linux is unhappy

csett86:
Indeed, prebuilds were not used during install, fixed it with

https://github.com/jitsi/robotjs/pull/7
https://github.com/jitsi/jitsi-meet-electron-sdk/pull/152

This PR is not urgent, obviously, I would wait for the next release of both deps until this then can be merged. Setting to draft until then.

saghul:
Approved both, makes sense!

csett86:
Now its ready as prebuilds are released (and used) here. As expected, build times on CI are also reduced a bit, thanks to the prebuilds.

saghul (approved):

PR #667, opened 2021-11-16
README: add publishing docs

csett86:


saghul (approved):
Thanks for doing this!

PR #668, opened 2021-11-17
refactor: Remove electron-store to be ready for electron 14+

csett86:
Release this 3 months after 2.9.0 was released (21 Sep 2021) so
that users had 3 months to migrate to local storage


saghul (approved):

saghul:
This is the last holdup for Electron 14, right?

csett86:
Yes this is the last holdup, but from a timeline perspective I suggest to wait another month to give users 3 months to migrate the settings

saghul:
Given the holiday season will be coming soon, how about we make this release now? (after SDK 3.0.2 is out, that is)

csett86:
Yes agree, electron 16 is also just out with a patch update to 16.0.3, and the contained Chromium 96 also seems to have no obvious regressions with jitsi (as its now out for 2 weeks as well and everything seems to work fine)

PR #669, opened 2021-11-24
mac: allow notarize with API key

csett86:
Use eg in GH actions like this with

secret API_KEY = API key file as coming from Apple
secret API_KEY_ID = API key ID as coming from Apple
secret API_KEY_ISSUER_ID = API key issuer ID as coming from Apple

    mkdir -p ~/private_keys/
    echo '${{ secrets.api_key }}' > ~/private_keys/AuthKey_${{ secrets.api_key_id }}.p8
    echo "API_KEY_FILE=~/private_keys/AuthKey_${{ secrets.api_key_id }}.p8" >> $GITHUB_ENV
    echo "API_KEY_ID=${{ secrets.api_key_id }}" >> $GITHUB_ENV
    echo "API_KEY_ISSUER_ID=${{ secrets.api_key_issuer_id }}" >> $GITHUB_ENV

### DIFF
--- a/notarize.js
+++ b/notarize.js
@@ -9,20 +9,34 @@ exports.default = async function notarizing(context) {
         return;
     }
 
-    if (!(process.env.APPLE_ID && process.env.APPLE_ID_PASSWORD && process.env.TEAM_ID)) {
-        console.log('Skipping notarization');
-
-        return;
-    }
-
     const appName = context.packager.appInfo.productFilename;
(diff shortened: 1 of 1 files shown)


csett86:
@saghul This does not break existing usage, just extending it so that notarization can also be done with API key instead of user / password credentials. Please review.

saghul (approved):
LGTM

PR #670, opened 2021-11-25
WIP to debug wrong arch of native node addons in mac universal build for x64

csett86:
- chore: package-lock.json noise
- 2021.11.3

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,6 +19,8 @@ jobs:
     - uses: actions/setup-node@v2
       with:
         node-version: '16'
+    - name: Install dependencies
+      run: sudo apt-get install libx11-dev zlib1g-dev libpng-dev libxtst-dev
     - name: Build it
       env:
         GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -19,6 +19,8 @@ jobs:
     - uses: actions/setup-node@v2
       with:
         node-version: '16'
+    - name: Install dependencies
+      run: sudo apt-get install libx11-dev zlib1g-dev libpng-dev libxtst-dev
```
saghul, line 23: Why is needed again?

csett86: We currently have the very unfortunate situation that local mac os builds on mac os 11 work fine for both archs, but the builds using the same prebuilds in gh actions fail to launch on intel macs. 

This pr is not meant to be merged, but I am using it to trying to get to the bottom of this issue and the provide a proper fix.

package.json
```diff
@@ -23,6 +23,7 @@
     "generateUpdatesFilesForAllChannels": true,
     "afterPack": "./linux-sandbox-fix.js",
     "afterSign": "./notarize.js",
+    "buildDependenciesFromSource": true,
```
csett86, line 26: This is also why I added this line to test how gh actions behave if prebuilds are disabled in electron-builder

csett86:
I will later add artifact archiving to the gh action so that I dont need to misuse a draft release to get to the build artifacts

.github/workflows/ci.yml
saghul: So the current release if broken on Intel Macs running macOS 11?

csett86: No, our current release works fine, as our current release predates the prebuild changes. But current master build with GH actions is broken on intel macs.

Downstream from ffmuc.net ran into this when releasing based on current master: https://github.com/freifunkMUC/jitsi-meet-electron/issues/38

saghul: Gotcha, thanks for looking into this!

PR #671, opened 2021-11-26
ci: archive binaries after build

csett86:


csett86:
This is now the archiving on each GH actions build. If you check the mac binaries and run them on x64, you will see the issue that ffmuc.net also ran into. Nontheless I think it would be great to have this merged also going forward. Do you agree @saghul?

saghul (approved):

PR #674, opened 2021-12-01
2021.12.1

csett86:
- 2021.12.0
- 2021.12.1

PR #675, opened 2021-12-01
deps: update electron-updater

saghul:
Fixes:
https://github.com/jitsi/jitsi-meet-electron/pull/674#issuecomment-983564898

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -40,14 +40,14 @@
         "babel-loader": "8.1.0",
         "concurrently": "5.1.0",
         "css-loader": "3.5.0",
-        "electron": "^16.0.3",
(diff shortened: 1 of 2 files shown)


csett86:
4.4.6 is not a stable release of electron-updater, so if we want to update I suggest 4.6.1 as this is current latest on https://www.npmjs.com/package/electron-updater

saghul:
oops

saghul:
Updated.

csett86:
Quick question on this: Did you test this with windows, mac and linux already? In addition, it currently conflicts...

saghul:
No I didn't, we can actually revisit later. I thought we needed it for the last release but we don't...

Issue #628, opened 2021-08-31
Debian Package Issue


oktay454:
There is a bug in the Debian package.
The files are located in the `/opt/Jitsi Meet/` directory.
However, when creating `/usr/bin/jitsi-meet` file in postinst script, it follows `/opt/jitsi-meet/` directory.
My humble advice, replace `/opt/Jitsi Meet/` with `/opt/jitsi-meet/`.

saghul:
This sounds like an electron-builder issue.

csett86:
Yes, this is an electron-builder issue fixed with https://github.com/electron-userland/electron-builder/pull/5941, which is part of electron-builder v22.11.7+

So this issue would be fixed with #583

csett86:
After further investigation: Its unfortunately not fixed with #583 alone, as the linux-sandbox-fix.js is not compatible with symlinks, and thus breaks the jitsi-meet command again :(

PR #684, opened 2021-12-04
chore: remove unnecessary debian 10 appimage fix

csett86:
Debian 11 (bullseye) changed the default for kernel.unprivileged_userns_clone to 1 (see https://www.debian.org/releases/bullseye/amd64/release-notes/ch-information.en.html#linux-user-namespaces).
Thus this workaround is no longer required, remove it.

This at the same time fixes the default command line entry for the deb package.

Closes: #628


csett86:
I tested this on Debian 11 and Fedora 35, all good. Ubuntu always had kernel.unprivileged_userns_clone to 1, so they were never affected in the first place and thus dont need this as well. Thus I suggest to drop this (and with that drop support for running the AppImage on Debian 10).

saghul (approved):

Issue #687, opened 2021-12-11
PreJoin Page does not update Username


JonnyTischbein:
* Tested with version: `2021-12-02` (appImage version)
* Tested on multiple jitsi instances (with enabled prejoin page): e.g. https://meet.jit.si https://meet.ffmuc.net (and some more)
* Tested on ubuntu 21.10 `5.13.0-22-generic x86_64 x86_64 x86_64 GNU/Linux`

The name, which is set in the Jitsi App Settings always overwrites the value entered in the prejoin page. 

Example: 
* Setting Name is set to `User A`
* On the PreJoin Page I enter `User B`
* I join as `User A` and also see in the console ` action Symbol(SET_NAME) @ 10:09:38.701 - app.js:173`

If I clear the Name set in the settings and join via a prejoin page I join without a username (even if I enter a name on the prejoin page).

I guess it is because of this line, where the app setting user name is set.

https://github.com/jitsi/jitsi-meet-electron/blob/22b34066132f7dd39920e7f3a2f3d3eb26695374/app/features/conference/components/Conference.js#L362-L379

When I join a conference and go to the "conference settings" -> profile and update there my username it is also updating the app setting username.

I guess because of this code:
https://github.com/jitsi/jitsi-meet-electron/blob/22b34066132f7dd39920e7f3a2f3d3eb26695374/app/features/conference/components/Conference.js#L157-L172

It seems to me that we miss an update of the username when it changes on the prejoin page

saghul:
Good analysis.

I have actually been thinking about something related this week: I think this app should remove those settings altogether and force the prejoin screen to be visible.

This way one could customize mucho more before actual joining, such as the devices, background, etc.

We have had prejoin for long enough that there would be no risk assuming it's supported, and we can easily override it from the iframe API.

A nice bonus is we'd delete a bunch of code from this project :-)

Thoughts @csett86 ?

csett86:
I agree that the prejoin page would be better than the current settings, as its very confusing to have both and they sometimes are in sync, sometimes not (as seen here).

So yes, forcing the prejoin via the iframe api sounds good!

saghul:
Excellent I'll try to take a crack at it.

Issue #22, opened 2018-03-25
No App Structure


AkshitKrNagpal:
Currently, the app does not have a structure. It would become difficult to understand and contribute to this project for others.
There should be separate folders for main processes, renderer processes and assets.

hristoterezov:
I think splitting the project into directories by functionality is better and more organized than splitting them into 2 folders - main and render. For example if we need to add some menus we can than create folder /menus, put the code related to the menus there and than use it in main.js or any other place that we need it.

I don't see what will be the benefit to have folders main and render.

Also the main process will always be 1, right? Am I missing something here?

AkshitKrNagpal:
@hristoterezov 
> Also the main process will always be 1, right? Am I missing something here?

The main process includes a lot of functions like creating and managing BrowserWindow instances, registering global shortcuts, dialogs, menus, system trays, notifications, etc. Auto updater and crash reporter are also a part of the main process. ( I am not saying we will be needing all right now or in future but definitely some)
Also, when dealing with multiple windows splitting current `main.js` into multiple files will make the code more organised.
It would be nice if we have a main folder where all this code can be placed. 


> I don't see what will be the benefit to have folders main and render.
Besides what I said above
* It will prevent new contributors to mix up main and renderer process code.
* It will also enable uniformity in code.


> I think splitting the project into directories by functionality is better and more organized than splitting them into 2 folders - main and render.

I don't disagree with you. But the benefits are good and the code inside these folders will be organised too.

saghul:
Let's close this and decide on every structure change as they are needed. As new features are implemented, we'll need changes, but it will become clearer once we get there.

PR #690, opened 2021-12-17
feat: force prejoin page instead of profile and av settings

csett86:
The settings and the prejoin page which is forced on many instances
including meet.jit.si did conflict, as you could set values in the
settings, then try to override them in den prejoin page only to notice
that it would not work. Clean this up by only having the prejoin page
where all settings can be done before joining the conference.

Closes: #687


csett86:
@saghul please review

csett86:
Need to fix the linting errors first, sorry for that

csett86:
Now its looking good, also tested that onboarding still works, and overall a lot simpler. As @atlaskit/panel is now no longer used anywhere, I removed that dependency as well.

JonnyTischbein:
I don't have that much experience how you review code in this project, but it looks good to me.

Thanks for your work :)

saghul:
Ah bummer, I had the same PR ready but forgot my laptop lol so I couldn't push it. Oh well, great to see it. I'll give it a read.

Not sure you did, but if you didn't, can you move the server URL setting to the non-advanced settings section now?

saghul (approved):
LGTM! I see you removed the advanced expansion panel, I thought of keeping it but I think it will just simplify things, so let's go with your approach!

JonnyTischbein:
Tested your branch on ubuntu 21.10 `5.13.0-22-generic #22-Ubuntu SMP Fri Nov 5 13:21:36 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`

#687 is solved for me. I also tested successfully changing the username via the `settings` after joining the conference.

PR #693, opened 2021-12-28
cleanup: insertable streams and wasm SIMD are on by default

csett86:
insertable streams since M86 according to https://chromestatus.com/feature/6321945865879552
wasm SIMD since M91 according to https://chromestatus.com/feature/6533147810332672
remote module is no longer used/present, thus remove that noop enable as well

### DIFF
--- a/main.js
+++ b/main.js
@@ -203,8 +203,7 @@ function createJitsiMeetWindow() {
         minHeight: 600,
         show: false,
         webPreferences: {
-            enableBlinkFeatures: 'RTCInsertableStreams,WebAssemblySimd,WebAssemblyCSP',
-            enableRemoteModule: true,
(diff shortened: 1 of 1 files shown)


saghul:
I keep getting emails to provide feedback for the SIMD origin trial.

Can you please check if SIMD is working without forcing it? I believe we log a line when you try to enable a virtual background.

csett86:
Unfortunately no logging, but:

Looks good that SIMD works without forcing it:

wasm-simd

csett86:
I now also found the logging, at least indirectly:

wasm-simd2

saghul (approved):

saghul:
Yep, looks good. Let's go!

PR #700, opened 2022-01-11
Revert "ci(mac): directly set env vars based on secrets"

csett86:
Reverts jitsi/jitsi-meet-electron#697

Unfortunately it breaks external PRs, see eg
https://github.com/jitsi/jitsi-meet-electron/runs/4774894331?check_suite_focus=true
for an example. Thus reverting it.


csett86:
@saghul can you please review this? I tested that it works in https://github.com/freifunkMUC/jitsi-meet-electron/pull/44

PR #706, opened 2022-02-08
feat: allow the jitsi-meet:// protocol in the URL box

csett86:
Allow this so that when you send around links to the app and someone
accidentally copy-pastes them into the URL box, things dont break.

### DIFF
--- a/app/features/conference/components/Conference.js
+++ b/app/features/conference/components/Conference.js
@@ -171,7 +171,7 @@ class Conference extends Component<Props, State> {
         const appProtocolSurplus = `${config.appProtocolPrefix}://`;
 
         // replace the custom url with https, otherwise new URL() raises 'Invalid URL'.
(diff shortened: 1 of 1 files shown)


app/features/conference/components/Conference.js
```diff
@@ -168,6 +168,12 @@ class Conference extends Component<Props, State> {
      * @returns {void}
      */
     _loadConference() {
+        const appProtocolSurplus = `${config.appProtocolPrefix}://`;
+
+        // replace the custom url with https, otherwise new URL() raises 'Invalid URL'.
+        if (this._conference.serverURL.indexOf(appProtocolSurplus) === 0) {
```
saghul, line 174: Meh check with startsWith? Since we only care about it in the start.

saghul (approved):

app/features/conference/components/Conference.js
csett86: Done

saghul (approved):

PR #711, opened 2022-02-15
chore(electron-bump) Update to electron 17

horymury:
- allow plan-b until we switch to unified plan


saghul (dismissed):
LGTM. Can you alto take it for a spin @csett86 ?

csett86:
I would directly update to electron 17.0.1 that was released yesterday, to include the patch updated chromium.

on the plan-b: Electron 16 contained M96 Chromium which also already did no longer support plan-b, thus: Any specific reason to reintroduce it now?

saghul:
> I would directly update to electron 17.0.1 that was released yesterday, to include the patch updated chromium.

Good catch!

> on the plan-b: Electron 16 contained M96 Chromium which also already did no longer support plan-b, thus: Any specific reason to reintroduce it now?

Jitsi Meet supports both and some (old) deployments may not have support for unified yet. What we are doing with the flag is to allow Electron to support it, not to use it necessarily.

horymury:
> I would directly update to electron 17.0.1 that was released yesterday, to include the patch updated chromium.
> 
> on the plan-b: Electron 16 contained M96 Chromium which also already did no longer support plan-b, thus: Any specific reason to reintroduce it now?


Changed to electron 17.0.1

PR #716, opened 2022-02-16
2022.2.1

csett86:


csett86:
I have tested this successfully under Windows 10, Ubuntu 20.04 and macOS 12

saghul (approved):

saghul:
Excellent, I just noticed and wondered if it was an oversight :-)

PR #717, opened 2022-02-17
ci(mac): fix sign & notarize setup

csett86:


csett86:
Sorry for the back-and-forth, its now working and I have tested it. This slipped through in #700

saghul (approved):

PR #728, opened 2022-03-09
chore(deps): update @jitsi/electron-sdk to fix aot issue

csett86:
See https://github.com/jitsi/jitsi-meet-electron-sdk/pull/200 for details


csett86:
Just a small bugfix update, please review @saghul. Afterwards I would cut the March release of this app, see #729

saghul (approved):

Issue #576, opened 2021-05-07
[Enhancement] APT repository for jitsi-meet-electron?


taivlam:
Would it be possible to offer this Jitsi desktop client via some APT repository (or have preexisting Jitsi repos include `jitsi-meet-electron`) for easier installation?  I encountered some struggles before even discovering the official `jitsi-meet-electron` desktop client for a Debian machine and didn't want others to also struggle with the same stumbling blocks.

* I am aware that install it via the AppImage or I can manually install the `.deb` package from this repository's Releases on Debian-based Linux distros using `sudo gdebi jitsi-meet-amd64.deb`.  I avoid the former because I prefer not to use AppImages.  However, the latter isn't entirely perfect since the `.deb` package isn't configured to "auto-update" `jitsi-meet-electron` via APT.  I could keep track of releases on GitHub via RSS and manually uninstall the old version to install the new version, though I don't see this method "scaling" very well if I have 2 or more Debian machines, even for non-work purposes.

* Additionally, I initially struggled figuring out where to start for installing a Jitsi Meet desktop client on Debian-based distros.  I essentially stumbled upon the desktop Jitsi Meet client by accident while browsing the AUR and found the `jitsi-meet-desktop` and `jitsi-meet-electron` entries.  (I use the former but not the latter on Arch Linux-based distros.)  Yes, the AUR is community contributed/unofficial, strictly speaking - however, the ease of installing via the AUR contrasts with my experience on Debian.

After encountering very outdated Debian repos mentioned in #331, I arrived to `jitsi-meet-electron` here on GitHub - mostly due to the Download page on https://jitsi.org/ not mentioning this client, as described in #265.

* While web searching, I nearly installed [the server software `jitsi-meet`](https://github.com/jitsi/jitsi-meet) - the [stable](https://jitsi.org/downloads/ubuntu-debian-installations-instructions/) version to be exact - which is not what I really wanted after spending a few minutes considering what the instructions actually achieved.

* Also I am aware Jitsi Meet "just works" (best on) Chromium-based browsers and is intended to be used in the browser.  However, the one feature that the desktop client has is the "Always on Top Window" - the little floating window when you navigate away from the main client window, similar to a feature the Skype desktop client also has - which makes the Jitsi Meet desktop client better than Jitsi Meet in the browser.

saghul:
This project is mostly maintained out of love. It can't get much time allocated to it, so we mostly implement low-maintenance stuff.

[...]

PR #734, opened 2022-03-21
debian: prepare for https://download.jitsi.org/stable/ repo

csett86:
Now requires fpm to be available in path, as the electron-builder
packages fpm version is too old for --deb-generate-changes parameter


csett86:
@damencho please review if this approach is fine for you and fits to the requirements to get the deb package included in the [https://download.jitsi.org/stable/ repo? Thank you in advance!

build-debian-meta.js
```diff
@@ -0,0 +1,43 @@
+/* global process */
```
saghul, line 1: just require process?

debian/jitsi-meet-electron.buildinfo
```diff
@@ -0,0 +1,163 @@
...
+ diffutils (= 1:3.7-3),
+ dpkg (= 1.19.7ubuntu3),
+ dpkg-dev (= 1.19.7ubuntu3),
+ dwz (= 0.13-5),
+ file (= 1:5.38-4),
+ findutils (= 4.7.0-1ubuntu1),
+ g++ (= 4:9.3.0-1ubuntu2),
+ g++-9 (= 9.4.0-1ubuntu1~20.04),
+ gcc (= 4:9.3.0-1ubuntu2),
+ gcc-10-base (= 10.3.0-1ubuntu1~20.04),
```
saghul, line 50: Where do all these dependencies come from? Also why the multiple flavours of gcc?

debian/jitsi-meet-electron.buildinfo
```diff
@@ -0,0 +1,163 @@
...
+Architecture: amd64
+Version: __VERSION__
+Checksums-Md5:
+ __MD5__ __SIZE__ __FILE__
+Checksums-Sha1:
+ __SHA1__ __SIZE__ __FILE__
+Checksums-Sha256:
+ __SHA256__ __SIZE__ __FILE__
+Build-Origin: Ubuntu
+Build-Architecture: amd64
```
saghul, line 13: Out of curiosity, do you how things would look like if we'd want to do arm64 builds?

.github/workflows/ci.yml
```diff
@@ -29,7 +32,9 @@ jobs:
       with:
         name: linux-binaries
         path: |
-          dist/jitsi-meet-amd64.deb
+          dist/*.deb
+          dist/*.changes
+          dist/*.buildinfo
           dist/jitsi-meet-x86_64.AppImage
```
damencho, line 38: So these will be uploaded as artifacts and we can use the links from the release to push it ... then we can trigger a remote ci job via web request, I think - the link and credentials will come from the env variables ... 
Can we trigger the jenkins job once the artifacts are uploaded and publicly available?

saghul: I think that should be doable from the ci.yml file yeah.

debian/jitsi-meet-electron.buildinfo
csett86: Then the arch should be a parameter, and the post-build step would need to iterate over both debs. I suggest that we do this once we really provide arm64 builds, as that anyway requires some refactorings. But if you think otherwise I can also directly add it.

debian/jitsi-meet-electron.buildinfo
csett86: These build-"deps" are a list of the build-essential packages that are installed in the build env. In this case I took a local snapshot, as recreating the perl script dpkg-genbuildinfo is quite some effort (and as far as I understood @damencho in #576 the .buildinfo file is not really used, thus I spared the effort and only provided the right files and checksums)

Apart from that, the long list is normal for the buildinfo file (as it just reflects what is installed on the system at the time, not what is really used for this specific build). See eg the buildinfo of a current jicofo, that also lists multiple gcc versions: https://download.jitsi.org/stable/jicofo_1.0-862-1_all.buildinfo

build-debian-meta.js
csett86: Yes, good idea, done

csett86: I also directly removed all other global process instances via #741

taivlam:
@csett86 Has there been progress on this pull request?  (The bot automatically marked this as stale, then closed it.)

csett86:
No, there is no progress on this in the last 2.5 years, thus I let the stale bot close it.

PR #738, opened 2022-03-24
chore(deps): remove process polyfill

csett86:
The @atlaskit/navigation also works fine without reference
to whether its running on mac or not (for the way its used in
this project).


csett86:
@saghul please review, I have tested it successfully on macOS

saghul (approved):

PR #739, opened 2022-03-24
chore: remove unused alias for process in webpack.main config

csett86:
process is available in the electron-main target, no need to alias

### DIFF
--- a/webpack.main.js
+++ b/webpack.main.js
@@ -19,10 +19,7 @@ module.exports = {
     resolve: {
         modules: [
             path.resolve('./node_modules')
-        ],
-        alias: {
-            process: 'process/browser'
-        }
+        ]
     }
 };
 


saghul (approved):

kb-0311:
Hello @csett86 . I had a few casual questions regarding this commit and the last one. My PR before was broken before I added those aliases and the process plugins. The app became functional only after I added the them.

Could you please just briefly explain how is the process getting resolved globally even when the alias' has been removed in main and render configs? 

Thanks.

csett86:
@kb-0311 For the main process the process module is always there and your polyfill addition was not required. 

On the renderer I removed the usage of process in #738 so that the alias was then no longer required. Thus yes, it was required previously but no longer after #738

kb-0311:
Cool, got it . Thanks a lot for your reply and time. 👍

PR #747, opened 2022-04-26
chore(external-api) Update external api script

horymury:
Fixes https://github.com/jitsi/jitsi-meet/issues/11416


csett86:
2 quick questions @horymury 

- Is this the version from current stable (2.0.7210)?
- Have you tested this successfully with older (eg. last stable) and current jitsi servers?

saghul:
This fixes a problem with current stable indeed. We inadvertently broke backwards compatibility.

This has since been fixed, but stable is affected.

saghul (approved):

csett86:
I understand that this fixes https://github.com/jitsi/jitsi-meet/issues/11416, but does it break for older jitsi meet then? I would then like to point it out in the April release (which is anyway overdue and would cut after merging this)

saghul:
Correct, this is ahead of stable because stable lacks a check that makes external API break.

We did a 2 part fix: in the al  external API itself, so it will work with old and new stable releases, and in the Jitsi Meet side, so the next stable will work with the old external API, this restoring backwards compatibility.

Good point on the commit though, @horymury can you please mention that in the commit description?

csett86:
As soon as thats done I would cut the release #750

horymury:
> As soon as thats done I would cut the release #750

Hi @csett86 , I re-updated the external_api.js from latest master, this time I added the commit tags and SH in the commit message.

saghul (approved):

PR #758, opened 2022-06-23
chore: Upgrade CI Dependencies to Latest Version

RudraSen2:
Upgrading CI Dependencies to Latest Version. Tested and Verified.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ on:
 jobs:
   build-linux:
     name: Linux
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


saghul:
I don't to run on `-latest` OS releases, because an upgrade may break them. I'd rather be explicit. WDYT @csett86 ?

csett86:
I agree @saghul, I also like reproducible dependencies and build environments, so explicit versions instead of -latest

csett86 (changes requested):
For all "runs-on", please use explicit versions instead of -latest to have reproducible dependencies and build environments.

RudraSen2 (commented):
Requested Changes Fulfilled.

csett86 (approved):

csett86:
Thanks, @RudraSen2

PR #759, opened 2022-06-23
chore(deps): electron-builder and electron-updater

csett86:
This is a prerequisite for universal mac builds with electron 18+
For details see https://github.com/electron-userland/electron-builder/pull/6750


csett86:
I have successfully tested this (including updating) on windows, linux and mac

csett86:
And once this is in, I would do the electron 17 -> 19 upgrade (containing chromium M102) and then cut the june monthly release

saghul (approved):

Issue #683, opened 2021-12-04
No Titlebar on Gnome-Wayland run natively


9Lukas5:
## Description

Running Jitsi in Wayland native on Gnome doesn't have Window decorator, as Gnome-Wayland doesn't support this at the moment.
This is related to the electron upstream issue electron/electron#27522 , but no idea how long it would take the upstream issue get's closed and then taking effect for Jitsi.

---

## Current behavior

Running Jitsi in Wayland native under Gnome results in the window not having a titlebar.
Possible to fully use and move around with Super+Mouse Click&Drag.

---

## Expected Behavior

Having Window decorator on Gnome when running in Wayland native

---

## Possible Solution

Add custom decorator, or wait what electron does upstream

---

## Steps to reproduce

Run in a Gnome-Wayland session with 

```
<path to bin> --enable-features=UseOzonePlatform --ozone-platform=wayland
```

---

# Environment details

Ubuntu 20.04 Focal Fossa
Gnome 41 on Wayland
Jitsi 2021.11.2

---

saghul:
This is not something we can solve, and having this issue open while Electron itself doesn't support won't help either.

When electron gets it we'll update.

9Lukas5:
Well, can not is maybe the wrong wording, as for example VSCode has an setting to enable client side decorations, as I found out [here](https://github.com/microsoft/vscode/issues/109176#issuecomment-833744341).

So the question would be more of is it worth the work, implementing something or waiting on what/when it changes in electron upstream^^

Jitsi doesn't have any menus, so it would basically only a simple bar, holding the buttons for minimize, maximize and close.

csett86:
Upstream electron is nearly ready with client side decorations, so I suggest that we hold off here and wait for that: https://github.com/electron/electron/pull/29618

snaggen:
The electron PR is merged, but I still do not see any borders when using native wayland....

saghul:
Try passing this argument: `--enable-features=WaylandWindowDecorations`

snaggen:
That worked! Thanks!

saghul:
@csett86 Not sure how risky it would be to enable this by default...

csett86:
Enabling by default at least under Ubuntu 22.04 (with wayland) causes issues, so I'd rather not do that by default:

![grafik](https://user-images.githubusercontent.com/1392875/177044802-51778ec9-651a-4602-ac0e-3a640dc1128d.png)

PR #766, opened 2022-07-03
feat(wayland): enable titlebar when running under wayland

csett86:
Use with --ozone-platform-hint=auto to enable

Closes: #683

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,6 +174,14 @@ To fix this, install libfuse2 as follows:
 sudo apt install libfuse2
 ```
 
+Under wayland, experimental native wayland support can be enabled with the command-line switch `--ozone-platform-hint` set to `auto`:
+
+```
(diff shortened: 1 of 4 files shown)


csett86:
This does not look right under Wayland (Ubuntu 22.04), so please don't merge for now. Will convert to a draft as well.

grafik

csett86:
I have reported the wayland titlebar issue upstream as https://github.com/electron/electron/issues/34820

csett86:
Now looking good after the electron update that contains the fixes:

image

csett86:
@saghul please review, I tested it successfully on Windows 10, macOS 12.5 and Ubuntu 22.04

csett86:
After this I would cut the August release: #783

saghul (approved):

PR #779, opened 2022-08-05
fix(ci): downgrade to windows-2019 for node-gyp

csett86:
The node-gyp bundled with node 16 does not yet support windows-2022.
Downgrade ci to windows-2019 for now.

For details see eg. https://github.com/newrelic/node-native-metrics/issues/182
and other projects like https://github.com/prebuild/prebuildify/commit/c6e4368f02b1c1965aec3b1f5564a8e15c239ba5

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,7 +66,7 @@ jobs:
           dist/latest-mac.yml
   build-windows:
     name: Windows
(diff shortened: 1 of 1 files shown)


csett86:
Does not fix it: https://github.com/jitsi/jitsi-meet-electron/runs/7695861348?check_suite_focus=true

PR #789, opened 2022-09-06
Add Chinese Translations

thttg:
Add Simplified Chinese and Traditional Chinese translations


app/i18n/index.js
```diff
@@ -12,9 +12,10 @@ const languages = {
     hu: { translation: require('./lang/hu.json') },
     it: { translation: require('./lang/it.json') },
     nl: { translation: require('./lang/nl.json') },
-    pt: { translation: require('./lang/pt-br.json') },
+    ptBR: { translation: require('./lang/ptBR.json') },
```
saghul, line 15: Why are you making this change?

thttg: https://github.com/jitsi/jitsi-meet/tree/master/lang

I think it would be better for both sides to use the same one.

"pt": "Portuguese",
"ptBR": "Portuguese (Brazil)",

csett86: Please dont put this unrelated changes into the chinese translation PR. 

In addition ptBR is not a proper language code, it would be pt-BR, as this is chosen by navigator.language, see https://developer.mozilla.org/en-US/docs/Web/API/Navigator/language

app/i18n/index.js
```diff
@@ -12,9 +12,10 @@ const languages = {
...
     ru: { translation: require('./lang/ru.json') },
-    sq: { translation: require('./lang/sq.json') }
+    sq: { translation: require('./lang/sq.json') },
+    zhCN: { translation: require('./lang/zhCN.json') }
```
csett86, line 18: zhCN is not a proper language code, please use zh-CN as this is compared with the return value of navigator.language, see https://developer.mozilla.org/en-US/docs/Web/API/Navigator/language

csett86 (changes requested):

app/i18n/index.js
thttg: Fixed

csett86 (changes requested):
Please run 

    npm run lint

locally and fix all reported issues

thttg:
image
I have no idea how to fix this one

csett86 (approved):
LGTM

Issue #285, opened 2020-04-18
Support X-Frame-Options DENY or ORIGIN


isnuryusuf:
Hi, 

is there any way to make jitsi-meet-electron support from X-Frame-Options Header Deny / sameorigin?
something like on this page:
https://github.com/electron/electron/pull/573 and https://github.com/electron/electron/issues/5036

really appreciate it your response

saghul:
Looks like this should work: https://github.com/electron/electron/pull/573#issuecomment-642216738

TheAssassin:
This issue has been open for a long time now. Some servers I want to use do set `X-Frame-Options` and thus break the electron app: it does not react any more, just shows the blue background (there is just an error message on the CLI which regular users won't see).

I'd like to see a workaround be implemented, but the one you suggested apparently strips the header completely, removing an important security feature. Perhaps it just needs to be replaced by `allow-from <local server>`?

saghul:
The Electron app's origin would be blank I think, so I'm not sure what we should be setting there TBH. If you resolve the problem a PR would be welcome.

PR #798, opened 2022-10-11
feat: allow connecting to instances that set CSP and x-frame-options

csett86:
While in browser environments the headers are sensible, the only purpose
of the electron app is load jitsi in the iframe api. Thus the security
implications are accepted to work with more jitsi instances.

Fixes: #285

### DIFF
--- a/main.js
+++ b/main.js
@@ -228,6 +228,17 @@ function createJitsiMeetWindow() {
 
     mainWindow.webContents.setWindowOpenHandler(windowOpenHandler);
 
+    // Filter out x-frame-options and CSP to allow loading jitsi via the iframe API
+    // Resolves https://github.com/jitsi/jitsi-meet-electron/issues/285
+    mainWindow.webContents.session.webRequest.onHeadersReceived((details, callback) => {
+        delete details.responseHeaders['x-frame-options'];
+        delete details.responseHeaders['content-security-policy'];
+
+        callback({
+            responseHeaders: details.responseHeaders
+        });
+    });
+
     initPopupsConfigurationMain(mainWindow);
     setupAlwaysOnTopMain(mainWindow, null, windowOpenHandler);
     setupPowerMonitorMain(mainWindow);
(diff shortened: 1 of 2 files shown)


main.js
```diff
@@ -228,6 +228,17 @@ function createJitsiMeetWindow() {
 
     mainWindow.webContents.setWindowOpenHandler(windowOpenHandler);
 
+    // Filter out x-frame-options and CSP to allow loading jitsi via the iframe API
+    // Resolves https://github.com/jitsi/jitsi-meet-electron/issues/285
+    mainWindow.webContents.session.webRequest.onHeadersReceived((details, callback) => {
+        delete details.responseHeaders['x-frame-options'];
+        delete details.responseHeaders['content-security-policy'];
```
saghul, line 235: I thought we had established that some CSPs do work?

saghul (dismissed):
The change LGTM (I did add a question) but I find the principle troublesome.

If a server admin sets X-Frame-Options they are being explicit about not wanting to be embedded. Browsers fulfill their wishes, so why are we ignoring them?

csett86:
My reasons are:

1. From a users perspective the app does not embed the conference, but loads it fullscreen and does not provide any additional functionality.

2. This would bring the mobile apps and the electron app in sync, as the mobile app also ignores these headers and loads conferences from servers that send these headers.

So for me the electron app should behave like the mobile apps and not like a browser here.

To reproduce: A public example of a server which sets those headers is https://sichere-videokonferenz.de/

saghul:
Makes sense, the mobile example is a good point.

Let's give it a go!

main.js
csett86: Good point, I changed the approach to only filter out the frame-ancestors CSP and keep everything else in place. With that, CSP would eg. still protect against loading 3rd party resources in the iframe itself if this is enforced via CSP.

csett86:
As I updated the handling of the CSP, could you please reapprove, @saghul?

saghul (approved):

PR #822, opened 2022-12-22
mac: remove unnecessary entitlements and cleanup notarize code

csett86:


package.json
```diff
@@ -173,7 +173,7 @@
     "electron-context-menu": "^2.5.0",
     "electron-is-dev": "^1.2.0",
     "electron-log": "^4.3.2",
-    "electron-notarize": "1.1.1",
+    "@electron/notarize": "1.2.3",
```
saghul, line 176: Alpha sort this please

entitlements.mac.plist
```diff
@@ -4,12 +4,6 @@
 <dict>
     <key>com.apple.security.cs.allow-jit</key>
     <true/>
-    <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
-    <true/>
-    <key>com.apple.security.cs.allow-dyld-environment-variables</key>
-    <true/>
-    <key>com.apple.security.cs.disable-library-validation</key>
```
saghul, line 11: IIRC this was necessary for certain virtual webcam softwares to work.

csett86: Thanks for the hint, I cross-checked and eg OBS requires it, I'll add it back

package.json
csett86: Sure, done

csett86:
@saghul please have another look and squash-merge if fine

csett86:
@saghul can you please take another look if its now fine for you?

Issue #829, opened 2023-01-17
Wayland: Screensharing is difficult to use


Alfagiovanni56:
*This Issue tracker is only for reporting bugs and tracking code related issues.*

Before posting, please make sure you check community.jitsi.org to see if the same or similar bugs have already been discussed. General questions, installation help, and feature requests can also be posted to community.jitsi.org.

## Description
I had installed the newest version of jitsi-meet-electron 23.1.0 ([jitsi-meet-amd64.deb) on my machine (Ubuntu 22.04). Within a wayland session - which is default on Ubuntu 22.04 - the app crashes whenever one tries screen sharing. I had a frozen image once, but most of the time the app simply closes down. I have also tried to install the latest AppImage, which causes exactly the same type of crash.
When opening an online session on Brave or Chrome, screen sharing is simply impossible within a Wayland session, with the single exception of other browser pages in Brave resp. Chrome. Finally even the Firefox (snap) crashes in a Wayland session when one tries to share a screen.
Unfortunately, this makes it almost impossible for me to use the latest versions of the electron app. Therefore I have decided to re-install the 22.12.0 version for the time being. I hope that it will be possible to reproduce the problem and to of course to resolved it as soon as possible.
My machine is an Ubuntu 22.04.1, Gnome 42.5 and Wayland.

## Current behavior
App crashes on screen sharing: closes down unexpectedly.

## Expected Behavior
Share one's screen going through the normal procedure / choices (whole screen or window etc...

## Possible Solution
???

## Steps to reproduce
Start session on ubuntu 22.04 / wayland and try screen sharing. Compare with an Xorg session 

# Environment details

![Schermata del 2023-01-18 00-53-17](https://user-images.githubusercontent.com/104439351/213037724-71eed3a6-4013-4b9d-9dce-c1b3126b55b3.png)

Alfagiovanni56:
I will try to do what you ask. I will have to re-install the 23.1.0 versions again, though. I'll do it as soon as I can, both with the deb and with the AppImage. As I wrote in my post, I had re-installed the 22.12.0 version, which does what it needs to do.

csett86:
[...]

PR #862, opened 2023-05-09
chore(deps) update Electron to 24

saghul:


saghul:
@csett86 WDYT about landing this and making a new release? We haven't done one in a bit...

csett86:
I'll take a look tomorrow. I was holding back the electron upgrade as it will break screensharing on wayland (segfault on trying), but as it does not seem to get fixed, I will add a bigger warning in the readme as a known issue together with this PR

saghul:
Thanks! Yeah that's unfortunate but 21 is EOLd so I think we have to upgrade.

csett86 (approved):
Looks good, tested successfully under mac OS 13, Windows 10. Linux works fine as well, apart from screensharing under wayland, which still segfaults. I will reopen #829 accordingly

Issue #870, opened 2023-05-26
 After upgrading to 2023.5.2 the app is showing white screen


cypherpork:
## Description
console log:
```
chromewebdata/:1 Not allowed to load local resource: file:///Applications/Jitsi%20Meet.app/Contents/Resources/app.asar/build/index.html
DevTools failed to load source map: Could not load content for chrome-error://chromewebdata/Applications/Jitsi Meet.app/Contents/Resources/app.asar/node_modules/conf/node_modules/ajv/dist/ajv.js.map: Unexpected end of JSON input
```
The second message is repeated.

When I followed the part to look for `.../build/index.html`, I saw that `app.asar` is not a directory but a file.

# Environment details
---
Macbook Pro M2 
Ventura 13.3.1 (a)

Windows 11 arm os shows the same behavior with a slightly different log:
```
chromewebdata/:1 Not allowed to load local resource: file:///C:/Users/miro/AppData/Local/Programs/jitsi-meet-electron/resources/app.asar/build/index.html
DevTools failed to load source map: Could not load content for chrome-error://chromewebdata/ajv.js.map: Unexpected end of JSON input
DevTools failed to load source map: Could not load content for chrome-error://chromewebdata/core.js.map: Unexpected end of JSON input
```
[chromewebdata-1685081873026-MacOS.log](https://github.com/jitsi/jitsi-meet-electron/files/11572224/chromewebdata-1685081873026-MacOS.log)

[chromewebdata-1685082637487-win.log](https://github.com/jitsi/jitsi-meet-electron/files/11572222/chromewebdata-1685082637487-win.log)

saghul:
Hum, maybe I messed up the check. It does run in dev mode, but release uses the asar bundle which might indeed be the problem.

Sorry about that, I'll take a look.

hellcry37:
yep same here

saghul:
Fix in https://github.com/jitsi/jitsi-meet-electron/pull/871 new release coming shortly.

Sorry everyone.

csett86:
Release is in progress (building currently), I will do a final test and then publish in about 30 mins

csett86:
Unfortunately with the fix from #871 its still not working for windows.

Can you please take another look, @saghul?

![Bildschirmfoto 2023-05-26 um 12 09 28](https://github.com/jitsi/jitsi-meet-electron/assets/1392875/b1d3e94c-1759-4b31-b4d9-f24db6d93172)

[...]

PR #871, opened 2023-05-26
fix(main) fix path check if it contains spaces

saghul:
They will be percent encoded and the check will fail.

Fixes: https://github.com/jitsi/jitsi-meet-electron/issues/870


csett86 (approved):
Tested successfully on macOS

PR #872, opened 2023-05-26
2023.5.3

saghul:


saghul:
Sorry @csett86 we need another one, I meeed it up :-/ we need this fix: https://github.com/jitsi/jitsi-meet-electron/pull/871

csett86 (approved):

PR #873, opened 2023-05-26
fix(main) fix path matching

saghul:
Use Node's helper to transform a file URL into a path, which works across platforms.

Fixes: https://github.com/jitsi/jitsi-meet-electron/issues/870


saghul:
@csett86  I tested Windows and macOS, I think we're good now 😅

csett86 (approved):
Looks good to me

Issue #883, opened 2023-07-13
No translations for Hindi [hi]


shellheim:
No translation available for the Hindi language. [IETF Code - hi]
I could make a pull request for that if no one's working on it currently, thanks.

saghul:
Feel free to do so! Thank you!

shellheim:
Hey! Thanks for allowing me contribute.  I just made the PR in #886, closed the previous #885 accidentally, sorry.

shellheim:
Closed by PR #888

PR #886, opened 2023-07-14
Add hindi translation

shellheim:
# Hey!  
This PR adds Hindi translation, addressing #883.
Thanks to @saghul for allowing me to contribute to jitsi!  
Also sorry for not making a draft PR at first.


app/i18n/index.js
```diff
@@ -1,36 +1,35 @@
-import i18n from 'i18next';
-import { initReactI18next } from 'react-i18next';
-import moment from 'moment';
+import i18n from "i18next";
```
saghul, line 1: Please do t change the existing style

app/i18n/index.js
```diff
@@ -1,36 +1,35 @@
...
+import { initReactI18next } from "react-i18next";
+import moment from "moment";
 
 const languages = {
-    de: { translation: require('./lang/de.json') },
```
saghul, line 6: Please don't change the existing style. Just add your changes with the same style as the surrounding code.

saghul (changes requested):

app/i18n/index.js
shellheim: I am so sorry, wasn't intentional. I was working with neovim and something about how it deals with copy/paste or something else entirely might have caused this, let me fix it.

shellheim:
It would be way more simple for me to do it from scratch then actually try to figure out git right now.

PR #889, opened 2023-07-16
ci: add support for linux-arm64

csett86:
Update electron-builder to be able to disable rebuilds, everything is prebuild anyway.
As part of electron-builder upgrade, fix sign and notarize on pull
requests for mac and continue to notarize outside of builder, because notarize within electron-builder
required teamId to be set in build config, but we want to keep that in env for now

See https://github.com/electron-userland/electron-builder/blob/master/packages/app-builder-lib/src/macPackager.ts#L525

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,10 @@ jobs:
         path: |
           dist/jitsi-meet-amd64.deb
           dist/jitsi-meet-x86_64.AppImage
+          dist/jitsi-meet-arm64.deb
+          dist/jitsi-meet-arm64.AppImage
           dist/latest-linux.yml
+          dist/latest-linux-arm64.yml
   build-mac:
     name: macOS
     runs-on: macos-11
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -32,6 +32,7 @@ jobs:
         path: |
           dist/jitsi-meet-amd64.deb
           dist/jitsi-meet-x86_64.AppImage
+          dist/jitsi-meet-arm64.AppImage
```
saghul, line 35: Shall we also add the deb?

csett86: Check if that works as cross-packaging under amd64

csett86: Local tests look good, I'll give it a try

csett86:
Its now ready for review, I have already tested it successfully on a Raspberry Pi with arm64 raspbian (bullseye). Due to the required electron-builder update I also did retests on Windows, Ubuntu and macOS, just to be sure that functionality and autoupdate still work as expected.

saghul (approved):
This is awesome!

PR #895, opened 2023-08-08
Enable the use of a static jwt token

d3473r:
Hi,

this PR enables the use of a static JWT, e.g. Server URL:  `https://meet.example.com?jwt=YOUR_JWT`
Our use-case is a meeting room with a device that should be able to join a JWT secured Jitsi instance for with other user get their JWT provided via Keycloak and https://github.com/d3473r/jitsi-keycloak
This device shouldn't have a user in our Keycloak server, but should be able to start meetings with a static JWT.

Regards

### DIFF
--- a/app/features/conference/components/Conference.js
+++ b/app/features/conference/components/Conference.js
@@ -190,7 +190,10 @@ class Conference extends Component<Props, State> {
         if (this._conference.serverURL.startsWith(appProtocolSurplus)) {
             this._conference.serverURL = this._conference.serverURL.replace(appProtocolSurplus, 'https://');
         }
-        const url = new URL(this._conference.room, this._conference.serverURL);
+
+        const serverUrl = new URL(this._conference.serverURL);
+        const jwt = new URLSearchParams(serverUrl.search).get('jwt');
+        const url = new URL(this._conference.room, new URL(serverUrl.pathname, serverUrl.origin));
         const roomName = url.pathname.split('/').pop();
         const host = this._conference.serverURL.replace(/https?:\/\//, '');
         const searchParameters = Object.fromEntries(url.searchParams);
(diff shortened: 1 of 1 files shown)


csett86:
@d3473r you need to help me with this: Even without the change this is already working for me, and this is also how meet.jit.si passes the jwt token as part of the URL towards the electron app. So I dont see why this additional code is required...

saghul:
Ditto, I didn't need to modify anything to make redirects to URLs with a token work...

d3473r:
Hmm, strange. For me this is not working as there is a url encoded trailing slash added to the url:

My steps to reproduce:

I configure the server url with the static JWT:
Screenshot 2023-09-05 at 11 24 07

Now try to create the room `test`

Screenshot 2023-09-05 at 11 21 40

Authentication is failing because of the appended slash on the JWT.
The inspected url: https://meet.example.com/test?jwt=xxx%2F
The trailing `%2F` is `/`

With this PR there is no trailing slash und the authentication is working as expected
Screenshot 2023-09-05 at 11 34 13

saghul:
Ah, hold on, you are assigning the token to the server URL! That is something very strange. If you are already using Keycloak why not use the token URL option to have the user be redirected to the auth server? That would be similar to how we do it with Firebase on meet.jit.si.

d3473r:
I want to use this for a conference room device, this device does not have a Keycloak user.
Otherwise everyone in the company who wants to create a meeting needs to known the SSO password for this device.
Also i don't want this device to have a SSO account, otherwise you could also login to different applications with it

d3473r:
We're using a Mac Mini as a Conference System with Teams and Jitsi on it.
Similar to a preconfigured Microsoft Teams Room Package: https://www.microsoft.com/en-us/microsoft-teams/across-devices/devices/category/teams-rooms/20

saghul:
Oh, that quite an unusual use case.

TBH I'm not sure what to do here. On the one hand supporting it seems easi enough (thought here should probably be a dedicated field in the advanced section instead of (ab)using URL parameters). But on the other it's so niche I doubt anyone else besides you needs this.

In addition, our physical rooms solution is Spot (https://github.com/jitsi/jitsi-meet-spot) so trying to make this app voerlap with Spot does not sound like a good idea to me.

d3473r:
Oh, i didn't know about spot. I took a look but don't understand yet how to integrate it within our current setup. But this is a separate issue :)
For now, I'll keep using a forked build from this repo

saghul:
Thanks for understanding!

PR #904, opened 2023-09-04
chore(deps) update electron to 26.1.0

csett86:
Contains the fix for the pipewire selector. Its now a combined selector, same as Chrome/Firefox, and contains the fix for the hard crash on wayland when stopping screenshare. Unfortunately sharing the full screen still does not work under wayland, only sharing individual windows (tested with Ubuntu 22.04).

Sharing a windows is still very cumbersome, as one needs to select the
window in the pipewire selector, then quickly select it in the jitsi
built-in selector as well and then it works. If one is too slow (>2
seconds) on the second selector, the pipewire selector comes up again.

See: #829

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -40,7 +40,7 @@
         "babel-loader": "^8.2.3",
         "concurrently": "5.1.0",
         "css-loader": "^6.7.1",
-        "electron": "25.3.0",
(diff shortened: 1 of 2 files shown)


appetrosyan (approved):
Please merge ASAP

saghul (approved):

PR #906, opened 2023-09-05
Fix recent list  to skip tokens

saghul:


saghul:
@csett86 PTAL. I made it so all entried will be normalized after a meeting is updated.

csett86 (approved):
Looks good to me, a quick test also worked well, thank you @saghul

damencho:
We got a question about this in the community call, is it possible to keep the config options and just drop the jwt?

saghul:
It should be possible, yeah.

Issue #908, opened 2023-09-07
v2023.9.0: Blank screen 


mindo:
### Description:

After the upgrade to [2023.9.0](https://github.com/jitsi/jitsi-meet-electron/releases/tag/v2023.9.0) I get a blank screen as soon as I click anywhere inside the jitsi window.

### Steps to reproduce:

1. Open jitsi
2. Click anywhere / wait a few seconds
3. Blank screen

### Expected behavior:

I should be able to connect to any server / room

### Actual behavior:

I get a blank screen.

### Server information:

This happens even before entering a url/room.

### Client information:

- Browser / app version: Appimage/deb 2023.9.0
- Operating System: Fedora 38 (wayland) + Kali Linux (x11)
![fedora_wayland_appimage](https://github.com/jitsi/jitsi-meet/assets/1084642/5f1259d8-3b5f-42a2-a62f-cd7b2f5e3a42)
![kali_x11_appimage](https://github.com/jitsi/jitsi-meet/assets/1084642/8c3ff3da-9bb3-40cd-ac55-32267582adeb)
![kali_x11_deb](https://github.com/jitsi/jitsi-meet/assets/1084642/ef05f660-bbfe-431e-9647-056dc39cbdec)

### Additional information:

Fedora logs:

`
➜ ./jitsi-meet-x86_64_bad.AppImage --enable-logging
Gtk-Message: 10:23:43.187: Failed to load module "appmenu-gtk-module"
10:23:43.356 › Checking for update
[1594135:0907/102343.518635:WARNING:sandbox_linux.cc(393)] InitializeSandbox() called with multiple threads in process gpu-process.
[1594044:0907/102343.537238:WARNING:bluez_dbus_manager.cc(247)] Floss manager not present, cannot set Floss enable/disable.
[1594044:0907/102344.129743:INFO:CONSOLE(2)] "%c action %cpersist/PERSIST %c@ 10:23:44.128 color: gray; font-weight: lighter; color: inherit; color: gray; font-weight: lighter;", source: file:///tmp/.mount_jitsi-a7mqej/resources/app.asar/build/app.js (2)
[1594044:0907/102344.129774:INFO:CONSOLE(2)] "%c prev state color: #9E9E9E; font-weight: bold [object Object]", source: file:///tmp/.mount_jitsi-a7mqej/resources/app.asar/build/app.js (2)
[...]

PR #910, opened 2023-09-13
fix(linux): downgrade electron to avoid renderer crashes

csett86:
Downgrade to electron 25 to avoid renderer crashes on Arch and Fedora.
See upstream https://github.com/electron/electron/issues/39775

Fixes: #908

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -40,7 +40,7 @@
         "babel-loader": "^8.2.3",
         "concurrently": "5.1.0",
         "css-loader": "^6.7.1",
-        "electron": "26.1.0",
+        "electron": "25.8.1",
         "electron-builder": "24.4.0",
         "electron-context-menu": "^2.5.0",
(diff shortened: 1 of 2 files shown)


csett86:
@saghul It seems this is failing on mac due to an expired mac certificate:

    "Developer ID Application: 8x8, Inc. (***)" (CSSMERR_TP_CERT_EXPIRED)

See https://github.com/jitsi/jitsi-meet-electron/actions/runs/6177176848/job/16767743589#step:5:890

Could I kindly ask you to provide the updated certificate in the secrets here in the repo?

saghul (approved):

saghul:
Sure I'll get one and drop a comment here.

saghul:
New key is ready, I reran the failed job and it's 💚  now!

csett86:
Thanks a lot @saghul

PR #923, opened 2023-11-14
chore: update sdk and external_api

gabiborlea:
updating electron sdk and external_api with screen sharing security fix


gabiborlea:
@csett86 if you can create a release after this is merged, would be great

saghul (approved):

gabiborlea:
yes

gabiborlea:
That one can be closed

PR #926, opened 2023-11-23
fix: add sandbox permission to iframe

andreicon:


saghul:
Change LGTM but I don't think it's enough for the download to happen. Have you tested this?

andreicon:
hi, saghul,

yes, i've tested this and the issue was fixed by adding that policy.

saghul:
How and where is the file actually downloaded?

andreicon:
The system's default "Save File" dialog is triggered by the web view. The user can then choose a location and specify a file name. I believe excalidraw specifies a mime type for the file, so the dialog already hints the file extension.

saghul:
That is pretty cool!

saghul (approved):

Issue #929, opened 2023-11-23
2023.11.2 requires at least jitsi-meet-web 1.0.7658-1, otherwise it returns to the main screen after 30sec


seregasoft:
## Description
2023.11.2 returns to the main screen after ~30sec after join

## Current behavior
v. 2023.11.2 sudden exit from the conference and returns to the main screen after ~30sec after join

## Expected Behavior
I continue to be in the conference

## Possible Solution
Workaround: use browser

## Steps to reproduce
Join to conference using jitsi-meet-electron v. 2023.11.2 
Wait 30 seconds

# Environment details
OS Windows 10
Client: jitsi-meet-electron v. 2023.11.2
Server:
jitsi-meet-prosody/stable,now 1.0.6644-1 all [installed,upgradable to: 1.0.7658-1]
jitsi-meet-turnserver/stable,now 1.0.6644-1 all [installed,upgradable to: 1.0.7658-1]
jitsi-meet-web-config/stable,now 1.0.6644-1 all [installed,upgradable to: 1.0.7658-1]
jitsi-meet-web/stable,now 1.0.6644-1 all [installed,upgradable to: 1.0.7658-1]
jitsi-meet/stable,now 2.0.7882-1 all [installed,upgradable to: 2.0.9111-1]
jitsi-videobridge2/stable,now 2.2-45-ge8b20f06-1 all [installed,upgradable to: 2.3-61-g814bffd6-1]

saghul:
Can you reproduce this on meet.jit.si ?

csett86:
I can reproduce it with jitsi-meet-web 1.0.7531-1, but not on meet.jit.si.

The timeout that this app uses (30 seconds by default) somehow fires, so the app no longer detects that it has joined the conference successfully... Nothing that I would find in the developer tools of electron so far...

Will check with current stable jitsi-meet-web 1.0.7658-1 as well.

csett86:
Good news, I cannot reproduce it with stable jitsi-meet-web 1.0.7658-1

saghul:
It's weird, however that it doesn't work with older versions. @gabiborlea can you PTAL tomorrow?

gabiborlea:
[...]

PR #931, opened 2023-11-24
fix(conference) fix not loading with older server versions

saghul:
Use an iframe event which is fired really early to detect if the iframe was loaded.

Incidentally the logic to detect if the iframe was actually loaded gets simplified now, since the only way for the event to be received is for the iframe to be loaded.

Fixes: https://github.com/jitsi/jitsi-meet-electron/issues/929


gabiborlea (approved):

saghul:
I'll make a new release now.

PR #943, opened 2024-01-21
dev: get electron-reload back to work

csett86:
we lost that somewhere as part of webpack upgrades I guess,
and I could not get it to work by configuring webpack differently,
but just removing the line in question works and the reload
also still works afterwards.

Signed-off-by: Christoph Settgast <csett86_git@quicksands.de>

### DIFF
--- a/patches/electron-reload+1.5.0.patch
+++ b/patches/electron-reload+1.5.0.patch
@@ -1,0 +1,13 @@
+diff --git a/node_modules/electron-reload/main.js b/node_modules/electron-reload/main.js
+index 688934d..860e071 100644
+--- a/node_modules/electron-reload/main.js
++++ b/node_modules/electron-reload/main.js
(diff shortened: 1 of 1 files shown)


saghul (approved):

saghul:
LGTM! Isn't this something to report to electron-reload?

csett86:
From what I read in upstream electron-reload issues, webpack issues dont seem to be on their priority list. But let me check again :)

csett86:
See here from the maintainer of electron-reload:

> The bottom line is: there is no valid use case for packing electron-reload with webpack.

Source: https://github.com/yan-foto/electron-reload/issues/90#issuecomment-687826933

Also electron-reload seems dormant, last commit is from July 2021, so I wont consider upstreaming for now

saghul:
Makes sense, thanks for giving it a try!

PR #957, opened 2024-04-07
fix npm vulnerabilities

h4sh5:
Bump versions of vulnerable components


saghul:
Did you just run npm audit fix?

Thanks for the heads up, we'll run it, but don't accept PRs like this because they are virtually impossible to audit.

PR #982, opened 2024-09-09
deps(mac): allow mac signing with password-less cert and update to current electron/notarize

csett86:


saghul:
Looks like we're good now!

PR #990, opened 2024-11-20
README: fix typos and grammar

taivlam:
* Use more consistent capitalization
* Reword slightly some phrases/sentences to sound more natural
* Use symbolic link to refer to GH issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,7 +161,7 @@ None
   ```
 
 * On Ubuntu 24.04 and later, the AppImage will fail with a sandboxing error (`The SUID sandbox helper binary was found, but is not configured correctly...`)
-  This is due to an AppArmor conflict that restricts unprivileged user namespaces (jitsi/jitsi-meet-electron#965,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -144,41 +144,41 @@ None
...
-  To fix this, install libfuse2 as follows:
+  To fix this, install `libfuse2` as follows:
 
   ```
   sudo apt install libfuse2
   ```
 
 * On Ubuntu 24.04 and later, the AppImage will fail with a sandboxing error (`The SUID sandbox helper binary was found, but is not configured correctly...`)
-  This is due to AppArmor conflicting by restricting unprivileged user namespaces (https://github.com/jitsi/jitsi-meet-electron/issues/965,
+  This is due to an AppArmor conflict that restricts unprivileged user namespaces (jitsi/jitsi-meet-electron#965,
```
saghul, line 164: I think the other link syntax is much more widespread. Please undo the change.

taivlam: Ok, I've made the change.

PR #995, opened 2025-01-13
fix: Adds support for passing jwt as has param.

damencho:


saghul (approved):

saghul:
@csett86 do you mind making a new release with this so we can remove the workarund from meet.jit.si? 🙏

csett86:
@saghul on it in #997

PR #1000, opened 2025-02-06
Add support for native getDisplayMedia

saghul:
(pending SDK bump)


saghul:
@csett86 Could you please land this and make a new release when you can? Note that the new screen-sharing flow will only work with Jitsi Meet as of today's master, which is not on meet.jit.si but it is on alpha.jitsi.net.

saghul:
nvm, I'm YOLO-ing it!

csett86:
Whohoo, thank you! Great to have the Wayland picker now directly as well

PR #1004, opened 2025-02-27
fix(macOS) fix missing thumbnails when enumerating SS sources

saghul:


main.js
```diff
@@ -35,9 +35,15 @@ const ENABLE_REMOTE_CONTROL = false;
 // We need this because of https://github.com/electron/electron/issues/18214
 app.commandLine.appendSwitch('disable-site-isolation-trials');
 
-// This allows BrowserWindow.setContentProtection(true) to work on macOS.
-// https://github.com/electron/electron/issues/19880
-app.commandLine.appendSwitch('disable-features', 'DesktopCaptureMacV2,IOSurfaceCapturer');
```
saghul, line 40: This is no longer necessary as the capturer has changed anyway.

quitrk (approved):

Issue #1011, opened 2025-03-29
Issue in Migrating `app/features/navbar/components/Logo`


naman9271:
I am trying to migrate a UI component in jitsi-meet-electron to TS where I got some errors while I am able to migrate non-UI code very easily

> [!NOTE]
> File Before migration 
```
import Icon from '@atlaskit/icon';
import React, { Component } from 'react';
import LogoSVG from '../../../images/logo.svg';

export default class Logo extends Component<*> {

    /**
     * Render function of component
     * @returns {ReactElement}
     */
    render() {
        return (
            <Icon
                glyph = { LogoSVG }
                size = 'xlarge' />
        );
    }
}
```
where i am facing issue due to atlaskit dependencies (for which i take help from [Discussion page of atlassian dev community](https://community.developer.atlassian.com/t/any-type-info-recommendations-available-when-using-atlaskit-with-typescript/13665)

![Image](https://github.com/user-attachments/assets/d3d11b1d-be5e-4a62-832e-c05b14862b53)
for which i have declared a `global.d.ts` which is 
![Image](https://github.com/user-attachments/assets/fa60b910-a45d-4741-b07c-7cc4a043be0d)

**So is this the correct approach for this ??**

And one more thing beside this error there are other very weird errors i got just after changing `.js` to `.ts`

![Image](https://github.com/user-attachments/assets/207b7d5a-e1fb-42cb-93a5-106fb2d8a232)

I tried defining interfaces inside the module declaration in `global.d.ts` to resolve the issue, but it ended up becoming messy. **Could you guide me on the correct approach to proceed?**

naman9271:
@saghul @damencho please guide me in this

saghul:
I think updating AtlasKit to a newer version might be a requirement before moving on to TS, since it's now written in TS and should have types.

naman9271:
yes @saghul I ahve updated the AtlasKit dependencies but still getting errors and i don't know why it is saying it is not declared when i am using this 

![Image](https://github.com/user-attachments/assets/a4174422-67cf-413f-aa26-dd0e2a648ec2)

and also for svg i have to declare this in `global.d.ts`
```
declare module '*.svg' {
    const content: string;
    export default content;
}
```

could u please tell where is the issue is ??

PR #1012, opened 2025-03-31
Updated AtlasKit dependencies

naman9271:
I have updated the atlaskit dependencies to their latest version please have a look!


saghul:
Are there really no API changes required?

naman9271:
Yes @saghul , I have only updated the Atlaskit dependencies to their latest versions without making any other code changes, in order to fix #1011. However, I noticed that the updated versions is not compatible with Jitsi Electron, which could require additional modifications. If there are any specific API changes needed, could you please tell me about them so that i can address them

saghul:
KNowing what needs to be changed is part of the update :-) We haven'tupdated them in years so of course breakage is expected.

naman9271:
Got it, My bad! I'll look into the errors and see what needs to be updated accordingly. Thanks for the heads-up

PR #1020, opened 2025-04-16
deps: update electron-builder and updater

csett86:
For the notarization options built into electron-builder,
see https://www.electron.build/app-builder-lib.interface.macconfiguration#notarize


csett86:
Still a draft, I want to test the updater first before this is merged

package.json
```diff
@@ -63,20 +61,6 @@
       "artifactName": "jitsi-meet-${arch}.${ext}",
       "category": "VideoConference;AudioVideo;Audio;Video;Network",
       "description": "Jitsi Meet Desktop App",
-      "desktop": {
```
saghul, line 66: Why are these no longer necessary?

package.json
```diff
@@ -21,7 +21,6 @@
     "appId": "org.jitsi.jitsi-meet",
     "productName": "Jitsi Meet",
     "generateUpdatesFilesForAllChannels": true,
-    "afterSign": "./notarize.js",
```
saghul, line 24: Oh, so it now works out of the box? Great!

csett86: Yes, it does :)

package.json
csett86: They are no longer supported in the new electron-builder, that's why I had to remove them

saghul (approved):
Great work!

PR #1021, opened 2025-04-28
feat(macOS) drop no longer needed entitlements

saghul:
These were added to support 3rd party (unsigned) virtual camera plugins, but with Apple tightening the delivery of unsigned code we shouldn't really open that door anymore.

### DIFF
--- a/entitlements.mac.plist
+++ b/entitlements.mac.plist
@@ -6,8 +6,6 @@
     <true/>
     <key>com.apple.security.device.camera</key>
     <true/>
-    <key>com.apple.security.device.microphone</key>
-    <true/>
     <key>com.apple.security.device.audio-input</key>
     <true/>
 </dict>
--- a/entitlements.mac.plist
+++ b/entitlements.mac.plist
@@ -4,8 +4,6 @@
 <dict>
     <key>com.apple.security.cs.allow-jit</key>
     <true/>
-    <key>com.apple.security.cs.allow-unsigned-executable-memory</key>
-    <true/>
     <key>com.apple.security.device.camera</key>
     <true/>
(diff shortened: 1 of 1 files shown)


csett86:
I had similar efforts in #822, could you quickly clarify why dropping these two but keeping allow-unsigned-executable-memory?

saghul:
That one is for the V8 JIT to work.

csett86:
I thought that the allow-jit should cover that, looking at the default entitlements upstream: https://github.com/electron/osx-sign/blob/main/entitlements/default.darwin.plist

saghul:
Looking though https://github.com/electron/osx-sign/issues/188 I'm not sure it's needed in 2025... we could just give it a try. I'll update the PR.

csett86:
And to not only rely on SO, I also checked Apples docs, and https://developer.apple.com/documentation/bundleresources/entitlements/com.apple.security.device.microphone is only for Sandbox, while https://developer.apple.com/documentation/bundleresources/entitlements/com.apple.security.device.audio-input is for Hardened Runtime (aka signed and notarized apps outside the Mac App Store)

saghul:
I think some people were distrusting forks in the store, so maybe we could leave it for them?

csett86:
The Mac App Store builds use separate entitlement files anyway in https://github.com/jitsi/jitsi-meet-electron/blob/master/resources/entitlements.mas.plist, so no need to keep that in the normal entitlement file, I think

saghul:
Oh true that!

saghul:
Updated!

csett86 (approved):
Looks good, tested successfully with macOS 14 (Sonoma)

PR #1022, opened 2025-04-29
chore: use more defaults from electron-builder

csett86:
in addition, the audio-input is not relevant for mas app sandbox,
remove it there. See
https://developer.apple.com/documentation/bundleresources/entitlements/com.apple.security.device.audio-input


csett86:
FYI asarUnpack was fixed for node addons in https://github.com/electron-userland/electron-builder/pull/8392

saghul (approved):

PR #1026, opened 2025-07-03
Enable pipewire camera support

Ourouk:
The features is integrated on chrome since a long time (still disabled by default)
Will probably be needed going forward has desktop on Linux add indicator for the usage of camera and microphone and rely on Pipewire to know if it's used.
Not saying that it need to be merged right now bc I don't know how stable it is from upstream chrome. 
Happily having my desktop indicator on my desktop still :-D

### DIFF
--- a/main.js
+++ b/main.js
@@ -50,7 +50,7 @@ app.commandLine.appendSwitch('force-fieldtrials', 'WebRTC-Audio-Red-For-Opus/Ena
 
 // Wayland: Enable optional PipeWire support.
 if (!app.commandLine.hasSwitch('enable-features')) {
-    app.commandLine.appendSwitch('enable-features', 'WebRTCPipeWireCapturer');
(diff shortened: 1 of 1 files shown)


saghul (approved):
I wouldn't mind giving it a try.

OrionRandD:
@Ourouk
[Enable pipewire camera support for better integration with gnome and kde](https://github.com/jitsi/jitsi-meet-electron/pull/1026/commits/c33d37675a1c86c19d0bc1c1c484be47f63bd67e)

I do not use either gnome not kde, but xfce4 and Stumpwm...
This change in pipewire messed my jitsi-electron after latest upgrade... no camera and no audio any more :(

Ourouk:
Yeah, it's probably fixable, that was a bit of an experimental merge anyway. 
Should probably add a wayland detector or pipewire ... Before adding the option.

PR #1057, opened 2025-12-19
feat(sdk) update to SDK 9 and Electron 39

saghul:
SDK 9 contains a new implementation of the Always on Top window which
uses the native browser Picture-in-Picture support.

### DIFF
--- a/main.js
+++ b/main.js
@@ -3,7 +3,6 @@
 const {
     initPopupsConfigurationMain,
     getPopupTarget,
-    RemoteControlMain,
     setupPictureInPictureMain,
     setupRemoteControlMain,
     setupPowerMonitorMain,


csett86 (dismissed):
See inline

csett86:
> > @saghul I tested it with macOS 15.7 and there the always-on-top / new pip does not work. Other than that, please fix the one nit from the linter, and I would still merge it anyway. If the pip is not working, I would accept that for the time being
> 
> What server did you test? If self-hosted, are you running the latest stable?

Good hint, now retested on latest jitsi-meet stable, but still no luck. I also checked the devtools console from electron, unfortunately nothing visible there

main.js
```diff
@@ -4,7 +4,8 @@ const {
     initPopupsConfigurationMain,
     getPopupTarget,
     RemoteControlMain,
```
csett86:
```suggestion
```

saghul:
Thank you!

PR #1074, opened 2026-02-09
fix: pass windowOpenHandler to initPopupsConfigurationMain

### LINKED ISSUE #1073: Authentication popup no longer opens browser after SDK 9 update
## Description

Since the 2026.1.0 release (commit 404660d, SDK 9 + Electron 39 update), clicking the authenticate button in a Jitsi Meet conference no longer opens the browser for authentication. The authentication flow for moderator login is completely broken — pressing the button has no effect.

## Root Cause

In the SDK 9 migration, `initPopupsConfigurationMain(mainWindow)` was introduced at `main.js:324`. This function internally calls `setWindowOpenHandler()` on the main window's webContents, which **overwrites** the previously registered `windowOpenHandler` (set at the former `main.js:242`).

The SDK's internal handler only allows Google and Dropbox OAuth popups. All other `window.open` requests — including the moderator authentication flow — fall through to a default `{ action: 'deny' }`, silently blocking the browser from opening.

`initPopupsConfigurationMain` accepts an optional second parameter `externalWindowOpenHandler` to serve as a fallback for non-OAuth popups, but it was not being passed.

## Steps to Reproduce

1. Start Jitsi Meet Electron 2026.1.0
2. Join a meeting on a server that requires authentication
3. Click the "Authenticate" button
4. **Expected:** Default browser opens for authentication
5. **Actual:** Nothing happens

## Affected Version

- 2026.1.0 (introduced in #1057)

## Fix

Pass the existing `windowOpenHandler` as the second argument to `initPopupsConfigurationMain()` so non-OAuth popup requests are properly delegated instead of silently denied.

Is most likely the root cause to Issue [1071](https://github.com/jitsi/jitsi-meet-electron/issues/1071) aswell

mod242:
- Fix authentication popup not opening browser after SDK 9 update
- Pass windowOpenHandler as second argument to initPopupsConfigurationMain() so non-OAuth window.open requests are properly delegated
- Remove redundant setWindowOpenHandler() call that was being overwritten

Fixes #1073

### DIFF
--- a/main.js
+++ b/main.js
@@ -239,8 +239,6 @@ function createJitsiMeetWindow() {
     windowState.manage(mainWindow);
     mainWindow.loadURL(indexURL);
 
-    mainWindow.webContents.setWindowOpenHandler(windowOpenHandler);
-
     if (isDev) {
         mainWindow.webContents.session.clearCache();
     }
@@ -321,7 +319,7 @@ function createJitsiMeetWindow() {
         callback(true);
     });
 
-    initPopupsConfigurationMain(mainWindow);
+    initPopupsConfigurationMain(mainWindow, windowOpenHandler);
     setupPictureInPictureMain(mainWindow);
     setupPowerMonitorMain(mainWindow);
     setupScreenSharingMain(mainWindow, config.default.appName, pkgJson.build.appId);


Zappo-II:
Works on my machine...

PR #1076, opened 2026-02-13
mac: prepare desktop audio capture

csett86:
With https://github.com/electron/electron/pull/49717 the desktopCapturer
supports system audio for macOS 13 and above. Is was recently fixed
in https://github.com/electron/electron/pull/49740 for current
electron.

This is a preparational step, but:

This on its own does not allow audio for screenshare, as its currently
hard-coded to only show the "Share Audio" button on windows, see
https://github.com/jitsi/jitsi-meet/blob/master/react/features/screen-share/functions.ts#L34

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
       "extendInfo": {
         "NSCameraUsageDescription": "Jitsi Meet requires access to your camera in order to make video-calls.",
         "NSMicrophoneUsageDescription": "Jitsi Meet requires access to your microphone in order to make calls (audio/video).",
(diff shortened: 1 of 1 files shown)


csett86:
@saghul I have redone this now cleanly, please review and approve. Its only a prep step and does not enable audio desktop sharing yet, as the "Share Audio" button is hard-coded in the web-app to only show on windows. I'll do a followup PR once this lands and is in a released version

saghul (approved):

PR #1087, opened 2026-03-05
Added: two-window layout with redesigned Recent Meetings UI

aadiexii:
**Changes**

- Implement two-window architecture (launcher + meeting window)
- IPC communication between launcher and meeting windows
- Redesign Recent Meetings UI to match the web client:
  - Flat row layout with date/time and room name/duration columns
  - Click anywhere on a row to join (removed separate Join button)
  - Trash icon (same SVG as web) hidden by default, visible on hover
  - Scrollable container with custom scrollbar
  - Frosted panel background for visual separation
- Adjust Header and Body spacing for responsive layout

@saghul I have made the possible changes, you can go through it and let me know how I can improve it more.


aadiexii:
@saghul 
Found an edge case I had missed, the SDK setup (PiP, screen sharing, power monitor) was still targeting the launcher window (mainWindow), but in the two-window architecture the conference iframe only exists in the meeting window. This meant PiP wouldn't trigger when minimizing the meeting window.

**Fixed in the latest commit:**
-Moved setupPictureInPictureMain, setupScreenSharingMain, setupPowerMonitorMain to target meetingWindow
-Added redirect blocking (will-redirect) to the meeting window for security parity with the launcher

csett86 (changes requested):
Please make the new UI tests translatable (see inline) and translate them if you speak any of the translated languages, @aadiexii

csett86 (dismissed):
@aadiexii Thank you, and we are getting there. Two questions and still one unnecessary whitespace change, please see inline.

One more important question: Where did you test this?

app/i18n/lang/en.json
```diff
@@ -1,33 +1,36 @@
-{
```
saghul: Leave this like it was.

main.js
```diff
@@ -494,10 +504,109 @@ ipcMain.on('jitsi-open-url', (event, someUrl) => {
...
+            preload: path.resolve(basePath, './build/preload.js'),
+            sandbox: false
+        }
+    });
+
+    meetingWindow.loadURL(indexURL);
+
+    // Store conference — sent to meeting window once its renderer-ready fires,
+    // guaranteeing the navigate-to-conference listener is already registered.
+    pendingMeetingConference = conference;
```
saghul: can't you add a handler to meetingWindow here like the ready event and then trigger the navigation?

saghul (commented):
This is looking very good! Left some comments, please take a look.

aadiexii:
@saghul  I'm currently working on the requested changes and will update you soon. Thanks for your patience.

saghul:
Great! Take your time.

aadiexii:
I have completely separated the UI redesign into another branch to focus solely on the architecture here as requested! Also, I have ditched the React Router, fixed the hackish Redux actions by creating a dedicated addRecentListEntry action, and cleaned up the pendingMeetingConference IPC logic by utilizing did-finish-load. Check out the latest commit!

app/features/welcome/components/Welcome.js
```diff
@@ -203,7 +203,7 @@ class Welcome extends Component {
      * @returns {ReactElement}
      */
     _renderHeader() {
-        const locationState = this.props.location.state;
+        const locationState = this.props.location && this.props.location.state;
```
saghul: use optional chaining

main.js
```diff
@@ -442,7 +429,11 @@ app.on('second-instance', (event, commandLine) => {
 });
 
 app.on('window-all-closed', () => {
-    app.quit();
+    // Only quit when the launcher (mainWindow) is closed.
```
saghul: why do you need this check, since the event should only fire when all windows are closed?

main.js
```diff
@@ -494,10 +493,109 @@ ipcMain.on('jitsi-open-url', (event, someUrl) => {
...
+
+        return;
+    }
+
+    const basePath = isDev ? __dirname : app.getAppPath();
+    const indexURL = URL.format({
+        pathname: path.resolve(basePath, './build/index.html'),
+        protocol: 'file:',
+        slashes: true,
+        query: { role: 'meeting' }
```
saghul: As I think I mentioned in the previous review, I don't like this, feels a bit like a hack. Can we find a different way? Perhaps having to separate index.html pages or something?

saghul (commented):
This looks much better, thank you! It's almost there, the  thing I'd like to see changed is how we tell the window apart, I'd like to see how an alternative to that query parameter looks like.

Also, if we no longer use the router, can we drop the dependency?

aadiexii:
Thanks @saghul! I've updated the PR with all your feedback:

-Uninstalled react-router completely and removed it from Redux.
-Applied optional chaining (props.location?.state).
-Removed the redundant mainWindow === null wrapper inside window-all-closed.
-Dropped the ?role=meeting query hack. The architecture now natively splits into two separate Webpack HTML entries (index.html and meeting.html).

-Bonus Fix: Wired addRecentListEntry to explicitly dispatch continuously so recent list timestamps successfully update when re-joining a meeting!
Let me know what you think of the native dual-HTML approach!

app/features/app/components/MeetingApp.js
```diff
@@ -0,0 +1,78 @@
...
+class MeetingApp extends Component {
+    /**
+     * Initializes a new {@code MeetingApp} instance.
+     *
+     * @inheritdoc
+     */
+    constructor(props) {
+        super(props);
+
+        document.title = config.appName;
```
saghul, line 22: Let's put the canonical meeting URL (URL with no parameters) once we have it, to make it more obvious where one is connected, since we don't have a location bar in Electron.

app/features/app/components/MeetingApp.js
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     * Implements React's {@link Component#componentDidMount()}.
+     *
+     * @returns {void}
+     */
+    componentDidMount() {
+        // Meeting window: set conference data when main process sends it.
+        this._removeConferenceListener = window.jitsiNodeAPI.ipc.addListener(
+            'navigate-to-conference',
+            conference => {
```
saghul, line 38: Let's do some validation here? I know we'll only send valid stuff over IPC, I wonder if a bug allows running code in the renderer could cause trouble here.

app/features/app/components/MeetingApp.js
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     * Implements React's {@link Component#render()}.
+     *
+     * @inheritdoc
+     * @returns {ReactElement}
+     */
+    render() {
+        const { conference } = this.state;
+
+        // Wait for the IPC message to deliver the conference details.
```
saghul, line 83: Maybe use a Spinner instead?

app/features/app/components/MeetingApp.js
saghul: We can also set the document title here after parsing the URL.

app/features/conference/components/Conference.js
```diff
@@ -59,16 +58,9 @@ class Conference extends Component {
 
         this._loadConference();
 
-        // Set a timer for a timeout duration, if we haven't loaded the iframe by then,
-        // give up.
+        // Set a timer: if the iframe hasn't loaded by then, close the meeting window.
```
saghul: The timer was there because otherwise there was no way to close the meeting, since there was a single window. Now that we have 2, let's drop the timer altogether.

app/features/recent-list/actionTypes.js
```diff
@@ -7,3 +7,13 @@
...
+
+/**
+ * The type of (redux) action that is dispatched when a conference is added to the recent list.
+ *
+ * @type {
+ *     type: ADD_RECENT_LIST_ENTRY,
+ *     conference
+ * }
+ */
+export const ADD_RECENT_LIST_ENTRY = 'ADD_RECENT_LIST_ENTRY';
```
saghul: can you add this one to the top, thus sorting them alphabetically?

app/features/recent-list/actionTypes.js
```diff
@@ -7,3 +7,13 @@
 * }
 */
 export const CONFERENCE_REMOVED = 'CONFERENCE_REMOVED';
```
saghul: Let's rename this to REMOVE_RECENT_LIST_ENTRY, for symmetry

saghul (commented):
I can see the finish line! Great work so far! Left a couple of comments, please take a look!

aadiexii:
Thanks! I pushed an update addressing all points:

-MeetingApp: Added proper IPC payload validation, dynamic {serverURL}/{room} UI titles, and a smooth @atlaskit/spinner for loading.
-Conference.js: Completely removed the legacy _loadTimer iframe expiration logic.
-Action Types: Alphabetized the list and globally renamed CONFERENCE_REMOVED to the symmetrical REMOVE_RECENT_LIST_ENTRY.

Let me know how it looks!

webpack.renderer.js
```diff
@@ -7,14 +7,24 @@ module.exports = {
     // The renderer code rus in BrowserWindow without node support so we must
     // target a web platform.
     target: 'web',
-    entry: { app: './app/index.js' },
+    entry: {
```
saghul: This change is creating 2 bundles with 99% the same code, except they render a different component.

A way around this is for app JS to register a global function that allows you to decide which entry-point you want to render, and you do that in the HTML file (you'd need to use a different HTML file for each entry-point, which is ok).

You can borrow some inspiration from here: https://github.com/jitsi/jitsi-meet/blob/9a6babd70ce18abd93ca99da5d62004dc262664d/react/index.web.js#L68

saghul (commented):
One last bug to avoid duplicating the entire JS bundle and we are good to go.

aadiexii:
Thanks, @saghul ! 
I've pushed an update addressing the final bug:
-Deleted app/meeting.js entirely.
-Rewrote app/index.js to register a single global renderEntryPoint function instead of auto-rendering.
-Split the HTML templates natively into app/index.html and app/meeting.html, both loading the unified app.bundle.js and passively calling renderEntryPoint('APP') and renderEntryPoint('MEETING') on load!

The bundle feels much lighter now. Let me know if everything is good to go! Then we can implement the UI part also.

saghul (approved):

saghul:
Great work @aadiexii !

PR #1126, opened 2026-05-11
chore: update electron-sdk to latest

csett86:
removes unused transitive electron-store dependency


saghul (approved):

saghul:
Do you know why we depend on babel-runtime in that weird way?

csett86:
Yes, the override on babel-runtime is from #1117 to remove the ancient core-js transitive dependency

PR #1129, opened 2026-05-11
Replace atlaskit with simple components

csett86:


saghul:
Do you have screenshots of how things are looking?

csett86:
Its the first try and it does not look too bad I would say:


grafik

grafik

grafik

csett86:
I would still fix the background on the main input bar, and maybe tweak it to look a bit more like meet.jit.si (so less blue background), but other than that I think it would work

csett86:
And to simplify I would remove the onboarding tour flow completely, see #1131 

The app is anyway so simple, let the people use it instead of offering the tour.

csett86:
Here now the updated styles:

grafik

grafik

What do you think, @saghul

csett86:
I did not like the white version, now a version inspired by the iOS Jitsi app:

Bildschirmfoto 2026-05-12 um 09 15 03

Bildschirmfoto 2026-05-12 um 09 16 09

saghul:
> I did not like the white version, now a version inspired by the iOS Jitsi app:

Way better!

saghul:
I'd like us to explore making the main window tall and narrow, and show a single column for the recent list, which is scrollable. That way it will take less screen real state...

csett86:
So even more like the iOS version

image

csett86:
I'll explore the redesign in the coming days, I like the idea to simplify the UI, thanks @saghul

saghul:
Yes, something along those lines, no need for tabs though. Thanks for taking the time to do this @csett86 !

csett86:
@saghul what do you say:

image

csett86:
Now the spinner is visible as well, I think we are good to go. If you have any further suggestion @saghul please keep them coming!

saghul:
Looks great! How about we land this and iterate as necessary? It's already quite large.

saghul (approved):

csett86:
> Looks great! How about we land this and iterate as necessary? It's already quite large.

Agree, lets land this as is and then iterate before the next release

PR #1136, opened 2026-05-29
build: replace webpack with esbuild

saghul:
Replace the dual webpack setup (webpack.main.js + webpack.renderer.js)
with a single esbuild script (esbuild.js) that builds the main, preload
and renderer bundles.

- main/preload: platform node, cjs, target node24; externalizes electron
  and the runtime deps that ship in node_modules (@jitsi/electron-sdk,
  electron-debug, electron-reload), bundling the rest.
- renderer: platform browser, iife; JSX in .js compiled natively,
  process.env.NODE_ENV defined, global polyfilled via banner. IIFE (not
  ESM) is used because the renderer is loaded from disk via a plain
  <script> over file://, where module scripts hit CORS restrictions; the
  bundle is a single self-contained file so ESM offers no benefit.
- the renderer's Chromium target is derived from the pinned electron
  version via electron-to-chromium (electron 41.1.0 -> chrome146), so it
  tracks electron bumps with no manual upkeep; this replaces
  babel-preset-env's targets: { electron }.
- esbuild plugins replace the webpack loaders/plugins: svgr (@svgr/core,
  dimensions:false/expandProps:start), external-api (keeps the vendored
  UMD external_api.js usable, replacing noParse), and html (renders
  index.html/meeting.html and injects the link/script, replacing
  HtmlWebpackPlugin).

Adopt Node 24 (electron 41's bundled Node): bump .nvmrc to 24 and
engines.node to >=24.0.0; verified with npm ci + build + lint on Node 24.

main.js no longer relies on webpack's node.__dirname rewrite: it derives
the app root from the bundle location (path.resolve(__dirname, '..')).

SettingsButton.js switches mousetrap to a default import so the bindGlobal
method added by the side-effect plugin import is visible (a namespace
import snapshots its members before the plugin runs).

Drops webpack, webpack-cli, the webpack loaders, html-webpack-plugin,
@svgr/webpack and the webpack-only Babel packages; adds esbuild,
@svgr/core, @svgr/plugin-jsx and electron-to-chromium.


saghul:
I had the clanker go to town on this one. Since the environment is very narrow (a specific Electron version) esbuild is much simpler and faster.

csett86 (approved):
Crazy to see how quick esbuild is, thanks!

